PR #22, opened 2016-06-04
correcting toLowerCase() for path

Event-Horizon:
Change hadn't been made yet per issues on the actual vscode-chrome-debug , made change and removed toLowerCase.


roblourens:
Thanks for the contribution, @Event-Horizon! I think that to make this a truly case-insensitive comparison, both sides should be lowercased. I've seen vscode, Chrome, and webservers change the casing of URLs in different ways. And I think the comment is unnecessary, and make sure all tests and tslint checks pass.

roblourens:
I fixed this in 0.3.5

PR #50, opened 2016-06-21
Add support for address param in launch config that allows remote debugging

nojvek:


nojvek:
cc @roblourens @auchenberg

src/chrome/chromeConnection.ts
```diff
@@ -162,16 +162,17 @@ export class ChromeConnection {
     /**
      * Attach the websocket to the first available tab in the chrome instance with the given remote debugging port number.
      */
-    public attach(port: number, url?: string): Promise<void> {
+    public attach(port: number, url?: string, address?: string): Promise<void> {
```
auchenberg: Logically I'd expect tha arguments to be : Address, port and url.

This smells of an after-thought... just one more param, right? I'd rather make sure the arguments are ordered logically, bump the sem-major version and update the two consumers of this debugger (edge + chrome)

nojvek: I thought about it. But since port is required, the other two are optional.
Having port, address, url looks equally weird as well.

auchenberg: Good point!

Issue #46, opened 2016-06-14
Replace handwritten Chrome Protocol .d.ts with better system


roblourens:
Something like https://github.com/DickvdBrink/chrome-debug-protocol - generating a .d.ts from the package.json. Would have to replace the callbacks with Promises. Would be great if the build fails automatically when the protocol changes.

roblourens:
Or https://github.com/nojvek/chrome-remote-debug-interface!

auchenberg:
What about https://github.com/cyrus-and/chrome-remote-interface? Has support for auto-detecting the support `protocol.json` by doing a HTTP GET `/json/protocol` which is supported by Edge and Node.

roblourens:
Callback-based :/. Also, we have an extra need to implement the server-side, for adapters. I think that's part of @nojvek's motivation to build his own.

auchenberg:
We should keep the notion of adapters, etc our of the core, but simply aim to have a runtime agnostic client that works with any CDP compliant endpoint.

roblourens:
Right, we need something that reads the protocol.json and does the code and typings generation for both sides. This repo would use it for the client side, and another repo would use it for the server side, or both sides.

auchenberg:
I'm a bit unsure of what you mean by client and server side. 

Using a 3rd party library for the actual CDP connection makes sense, while exposing a vscode debugger abstraction for other debuggers in this library makes sense to me.

roblourens:
On one end of the connection, a client is sending `setBreakpoint` and receiving `scriptParsed` events, and on the other, a sever is receving `setBreakpoint` requests and sending `scriptParsed` events, and the code that we should generate for the two is different but closely related.

nojvek:
👍. I think the abstraction could be cleaner. I'll submit a PR.

PR #68, opened 2016-07-13
Add emulation commands to chrome connection

PatoBeltran:
Add emulation chrome commands which are helpful for viewport resizing.

### DIFF
--- a/src/chrome/chromeConnection.ts
+++ b/src/chrome/chromeConnection.ts
@@ -245,6 +245,32 @@ export class ChromeConnection {
         return this.sendMessage('Page.setOverlayMessage');
     }
 
+    public emulation_clearDeviceMetricsOverride(): Promise<Chrome.Response> {
+        return this.sendMessage('Emulation.clearDeviceMetricsOverride');
+    }
+
+    public emulation_setEmulatedMedia(media: string): Promise<Chrome.Response> {
+        return this.sendMessage('Emulation.setEmulatedMedia', { media });
+    }
+
+    public emulation_setTouchEmulationEnabled(enabled: boolean, configuration?: string): Promise<Chrome.Response> {
+        let messageData: { enabled: boolean; configuration?: string; } = { enabled };
+
+        if (configuration) {
+            messageData.configuration = configuration;
+        }
+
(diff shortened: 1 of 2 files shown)


auchenberg:
The purpose of the core debugger is not to provide a client library for the APIs supported for by Chrome and other Chrome Debugger Protocol compliant runtimes.

If this is needed I suggest that we could add a generic method to trigger commands on the connected CDP connection, @roblourens?

nojvek:
Who will be calling the functions? Aren't they going to be dead functions for time being?

roblourens:
I told Patricio to add these, to hold vscode-cordova over until we do something like #46

auchenberg:
I see this as a slippery slope. These should be added in the runtime specific debugger, and not in the core. The core should expose generic logic to make API requests.

roblourens:
`ChromeConnection` is sort of the generic logic to make API requests. It will be outsourced to some external module, which will be pretty easy because it's already totally separate from the actual logic in -core.

auchenberg:
And yet you just expanded the API surface exposed by `core` to include API Emulation features that will you will need to support in the future to be backwards compatible? 

I made my point, but you are the maintainer, so it's your call.

roblourens:
Well, I'm not concerned about backwards compatibility here, to be honest. This interface will break when we do #46 and replace ChromeConnection with some other library that provides the CDP endpoint. And that interface should break whenever the protocol changes.

nojvek:
That's right, the goal of crdi is to pull nightly changes from chromium
protocol jsons, generate an interface like this and publish to npm if any
changes.

Since it has both the adapter and client interfaces, the edge adapter can
implement the adapter interface and vscode / other clients can use the
client interface.

The request, response, event object interfaces are shared between client
and server.

e.g.

https://github.com/nojvek/chrome-remote-debug-interface/blob/master/src/crdi.ts

auchenberg:
Let me keep you in mind that the supported API is an dynamic target and depend on the target client, hence the added ability to fetch `protocol.json` on demand on other clients and in runtimes like Edge and Node. 

Relaying on the Chromium specific API would be a short term thing.

PR #75, opened 2016-07-29
add getProperty to show getter property value

f111fei:
description see https://github.com/Microsoft/vscode/pull/9949


roblourens:
I'll keep an eye on that issue, thanks for the PR. The Node adapter already resolves getters. Requiring a click is nice because there can be side effects, but in a getter, there shouldn't be.

f111fei:
yeah. `node-debug` trigger a getter by default.  
https://github.com/Microsoft/vscode-node-debug/blob/master/src/node/nodeDebug.ts#L2304

It's also possible to require a click by marked the variable is unsolved.

PR #119, opened 2016-10-14
Expose chrome connection events and clients to extending class

iradul:
hookConnectionEvents function will allow custom adapter class that
extends ChromeDebugAdapter to listen to events that are not handled by
base ChromeDebugAdapter class.

runConnection function will allow custom class to enable clients other
than Debugger and Runtime which are the only one currently enabled.

iradul/vscode-phantomjs-debug#5

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -224,6 +224,34 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         }
     }
 
+    /**
+     * Hook up all connection events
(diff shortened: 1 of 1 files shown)


roblourens:
Thanks for the PR! I agree that this is needed for subclassing, because you want to do this in that time between attach and firing 'initialized'.

I'm actually in the middle of changing this to use the typed events instead of 'on', so note that `chromeConnection.on` will go away in favor of stuff like this -
`this.chrome.Debugger.onPaused(params => this.onPaused(params));`.

PR #126, opened 2016-11-08
don't fall on empty params in resolveParams

jan-zajic:
sometimes i can debug our applications only with this fix


roblourens:
Thanks, but I'm guessing your problem is the one fixed by https://github.com/Microsoft/vscode-chrome-debug-core/commit/e3faf88ff186f00df28c4680baf5bb0682c0e76f. Can you try that? With this change instead, you'll miss console logs.

PR #147, opened 2017-01-05
add path mapping

### LINKED ISSUE #146: Map local folder to URL
If I check sourceMap spec, it specify where we found URL source 
...
sources – An array of URLs for the original source files.
sourceRoot – (optional) The URL which all of the files in the sources array will be resolved from.
...

I haven't found a way **to map sources url to source local path** like this (like in chrome when we **simply** map a local folder to network resource...)
- "${workspaceRoot}/project1/src/*" -> ["http://localhost:8080/myapp/project1/*", "https://my_remote_website_in_production/myapp/project1/*"]
- "${workspaceRoot}/project2/src/*" -> ["http://localhost:8080/myapp/project2/*", "https://my_remote_website_in_production/myapp/project2/*"]

sourceMapPathOverrides don't allow this kind of config and this is not a shared config over launchers.

llgcode:
Hi @roblourens 
this pr fixes #146 and have to be merged with Microsoft/vscode-chrome-debug#324
I added pathMapping in the launcher config:
So it replace webRoot and sourceMapPathOverrides that are no longer needed to configure a launcher
I've tested with some typescript and sourcemap and it seems to work fine with just a pathMapping.

It was tricky to configure the debug environment but I finally installed projects and found the right procedure for launching.

### DIFF
--- a/src/debugAdapterInterfaces.d.ts
+++ b/src/debugAdapterInterfaces.d.ts
@@ -15,6 +15,7 @@ export type ISourceMapPathOverrides = { [pattern: string]: string };
  */
 export interface ICommonRequestArgs {
     webRoot?: string;
+    pathMapping?: {[url: string]: string};
     outDir?: string;
     outFiles?: string[];
     sourceMaps?: boolean;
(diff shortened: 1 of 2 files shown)


msftgits:
Hi, I am closing and re-opening this PR to bump the CLA bot. Sorry for the inconvenience!

roblourens:
This is great, thanks for the PR! I don't think I want to deprecate webRoot, since it's simple and almost always the only thing someone needs. But I'll merge and test and release this when I have a minute.

PR #151, opened 2017-01-09
Support https protocol using any certificate

### LINKED ISSUE #65: Does not work if the url is under https
The code is not working correctly if the url protocol is `https` and there are source maps because it cannot download the source map file.

It looks like the problem is the usage of the `http` module in `utils. getURL`. A better alternative is the `fetch` API available via `node-fetch` module.

kanongil:
This fixes #65 by adding `https` support to `getURL()`.

Note that I always set the `rejectUnauthorized = false` option, which disables the certificate validation checks. This is done to enable fetching from self-signed certificates, which are often used for local https development setups. This should be safe as long as it the method is only used for local resources.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -9,6 +9,7 @@ import * as path from 'path';
 import * as glob from 'glob';
 import {Handles} from 'vscode-debugadapter';
 import * as http from 'http';
+import * as https from 'https';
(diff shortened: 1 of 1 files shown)


roblourens:
Awesome, thanks.

PR #190, opened 2017-03-22
Support source mapping of stack traces in the Debug Console #6

nojvek:
Will add some tests tomorrow. Still figuring out how tests work with sourcemap mocks.

Let me know if the code section looks good by you.

@auchenberg @roblourens

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -740,16 +740,16 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
     protected onExceptionThrown(params: Crdp.Runtime.ExceptionThrownEvent): void {
         const formattedException = formatExceptionDetails(params.exceptionDetails);
-        this.mapFormattedException(formattedException).then(formattedException => {
+        this.sourceMapFormattedException(formattedException).then(exceptionStr => {
             this._session.sendEvent(new OutputEvent(
-                formattedException,
+                exceptionStr,
                 'stderr'
             ));
         });
     }
 
     // We parse stack trace from `formattedException`, source map it and return a new string
-    protected async mapFormattedException(formattedException: string): Promise<string> {
+    protected async sourceMapFormattedException(formattedException: string): Promise<string> {
         const exceptionLines = formattedException.split(/\r?\n/);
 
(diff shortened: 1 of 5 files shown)


roblourens:
Thanks, this looks great! Tests are always welcome.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -740,10 +740,37 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
+        const exceptionLines = formattedException.split(/\r?\n/);
+
+        for (let i = 0, len = exceptionLines.length; i < len; ++i) {
+            const line = exceptionLines[i];
+            const matches = line.match(/^\s+at (.*?)\s*\(?([^ ]+\.js):(\d+):(\d+)\)?$/);
+
+            if (matches) {
+                const fnName = matches[1];
+                const path = matches[2];
+                const lineNum = parseInt(matches[3]);
```
auchenberg: TSLint failes due to missing radex param.

nojvek:
Just installed `tslint` plugin for vscode. Just realized that that it now has `tslint.autofix` like eslint. 

I press save and everything fixes itself. So awesome 😎 🎉

nojvek:
@roblourens I am trying to get the unit test for exceptions to work.

Firing `mockEventEmitter.emit('Runtime.exceptionThrown', ...` but that never ends up calling `onExceptionThrown` even thought the handler is run `     this.chrome.Runtime.onExceptionThrown(params => this.onExceptionThrown(params));
`. Weird, am I doing something obviously wrong that I can't see?

My test strategy is to fire a bunch of `Runtime.exceptionThrown` events, mock the sourcemapper and pathmapper and check the final exception string sent to client is source mapped.

roblourens:
Looks like there's a typo in my mock event handling code - https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/test/mocks/debugProtocolMocks.ts#L51 - should be `Runtime.exceptionThrown` not `onExceptionThrown`.

nojvek:
@roblourens Thanks for the pointer. 

Making mocks work properly was a mission but I think I have it all working now.

nojvek:
Tested on dummy and webpack projects. `I'm lovin it` 

image

roblourens:
Awesome, thanks for the tests

PR #194, opened 2017-04-03
Fixed a bug where pathMapping had to end with a slash.

mlewand:
This PR fixes issue Microsoft/vscode-chrome-debug#393.

There's one thing that I'm not especially proud of in [src/chrome/chromeUtils.ts](https://github.com/mlewand/vscode-chrome-debug-core/blob/7f7c119352c7374e16d2be3eaeaac7808c2f2d03/src/chrome/chromeUtils.ts#L29-L30) file. IMHO it would be better more readable just to go over `pathMapping` members and override it's keys by right-trimming slashes in keys. I'm not comfortable with modifying `pathMapping` object (someone might get surprised after using `targetUrlToClientPathByPathMappings`) so ultimately I'd just clone the `pathMapping` object first, and then trim keys in this local clone.

However I don't what are environment constraints of this plugin, ideally I'd simply use `Object.assign` (shallow clone is OK here), but it's ES6. Writing clone by in loops is not a fun thing to do, and adding a dep just for that seems to be overkill.

Let me know if you'd want me to correct it, or feel free to do it on top of my changes - this use case has now good coverage, so you're safe to modify it.

### DIFF
--- a/src/chrome/chromeUtils.ts
+++ b/src/chrome/chromeUtils.ts
@@ -24,13 +24,8 @@ export function targetUrlToClientPathByPathMappings(scriptUrl: string, pathMappi
             p = p + '/';
         }
 
-        let localPath = pathMapping[origin + p];
-        if (localPath) {
-            const r = decodeURIComponent(parsedUrl.pathname.substring(p.length));
-            return path.join(localPath, r);
-        }
+        let localPath = pathMapping[origin + p] || pathMapping[p];
 
-        localPath = pathMapping[p];
         if (localPath) {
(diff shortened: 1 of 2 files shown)


roblourens:
Thanks for the PR, and thanks for including unit tests! I think this approach is fine, honestly. The data is stored in the given format, and we query it by all the formats we support. That makes sense to me. `Object.assign` is ok, because the only environment I'm concerned about is the version of Node shipped with VS Code - currently 6.5, soon to be 7.4. So I can use most ES6 features.

PR #199, opened 2017-04-11
Prefer address in launch.json over what is returned by target

louiedp3:
This is to support having different IPs between a host and a container
Issue Microsoft/vscode#23257


louiedp3:
I wasn't sure if this was the appropriate place to put this translation. Please let me know if I should move it elsewhere.

Cheers

roblourens:
This looks perfect, thanks!

PR #201, opened 2017-05-08
Fix Microsoft/vscode-chrome-debug#426: handle double dots in path

bpesquet:
I used a regexp to handle double dots and updated the test suite accordingly. 
Linting and tests are ok.

### DIFF
--- a/test/chrome/chromeUtils.test.ts
+++ b/test/chrome/chromeUtils.test.ts
@@ -318,6 +318,20 @@ suite('ChromeUtils', () => {
                 chromeUtils.getMatchingTargets(targets, 'http://localhost'),
                 targets);
         });
+
+        test('handles double dots', () => {
+            const targets = makeTargets('http://localhost/app', 'http://localhost/site/../folder/../app');
+            assert.deepEqual(
+                chromeUtils.getMatchingTargets(targets, 'http://localhost/site/../folder/../app'),
+                [targets[0]]);
+        });
+
+        test('handles a series of double dots', () => {
+            const targets = makeTargets('http://localhost/app', 'http://localhost/site/folder/../../app');
+            assert.deepEqual(
+                chromeUtils.getMatchingTargets(targets, 'http://localhost/site/folder/../../app'),
+                [targets[0]]);
+        });
     });
 
     suite('compareVariableNames', () => {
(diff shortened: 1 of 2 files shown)


roblourens:
Thanks for the PR! It looks like this isn't resolving the .., but just removing it, which wouldn't fix the original issue, right? And I'd much rather use `path.resolve` from Node, since it will handle single dots and all the edge cases.

test/chrome/chromeUtils.test.ts
```diff
@@ -318,6 +318,27 @@ suite('ChromeUtils', () => {
                 chromeUtils.getMatchingTargets(targets, 'http://localhost'),
                 targets);
         });
+
+        test('handles double dots', () => {
+            const targets = makeTargets('http://localhost/app', 'http://localhost/site/../folder/../app');
```
roblourens: Thanks for writing tests, but the target won't have `..`, and these tests aren't actually testing that the `..` is resolved, right?

bpesquet:
> It looks like this isn't resolving the .., but just removing it, which wouldn't fix the original issue, right?

Actually it removed the `..` *and* the parent folder, so that `localhost/site/../app` becomes `localhost/app`.

Anyway, I followed your advice and used `path.normalize()` instead. My local tests run fine, but some CI ones on AppVeyor fail. I may have missed something, my apologies if it is the case.

roblourens:
I don't think we can apply path.normalize to everything, as it will break URLs:

```
> require('path').normalize('http://localhost')
'http:/localhost'
```

I'm not sure we should resolve .. in all URLs anyway. Just `file:///` URLs.

roblourens:
I don't think we can apply path.normalize to everything, as it will break URLs:

```
> require('path').normalize('http://localhost')
'http:/localhost'
```

I'm not sure we should resolve .. in all URLs anyway. Just `file:///` URLs.

roblourens:
Appveyor runs on windows, and if you send a URL through a `path` API without fixing it afterwards, you'll get backslashes back.

test/chrome/chromeUtils.test.ts
bpesquet: Sorry, I don't understand what you mean by "resolve". A double dot in a path means going up one level in the folder structure: `my/path/to/this/../resource` resolves to `my/path/to/resource`. 

That's precisely my use case (`file:///Users/baptiste/Documents/Projets/GitHub/bpesquet/thejsway-code/chapter02/js/../html/celsius_fahrenheit.html` should resolve to `file:///Users/baptiste/Documents/Projets/GitHub/bpesquet/thejsway-code/chapter02/html/celsius_fahrenheit.html`), and I think that's what my code does.

roblourens: Actually I see, this is testing that it matches `targets[0]` and doesn't match the one with dots. Sorry, I misread it.

PR #214, opened 2017-07-03
Added parameter to multiplex the CRDP connection into an additional endpoint

digeff:
The multiplexer will let us attach other components that use the CRDP protocol to the debugger, such as a DOM Explorer.

### DIFF
--- a/src/chrome/chromeConnection.ts
+++ b/src/chrome/chromeConnection.ts
@@ -13,6 +13,9 @@ import {ChromeTargetDiscovery} from './chromeTargetDiscoveryStrategy';
 import {Client} from 'noice-json-rpc';
 import Crdp from '../../crdp/crdp';
 
+import {CRDPMultiplexor} from './crdpMultiplexing/crdpMultiplexor';
+import {WebSocketToLikeSocketProxy} from './crdpMultiplexing/webSocketToLikeSocketProxy';
+
 export interface ITarget {
     description: string;
     devtoolsFrontendUrl: string;
@@ -82,6 +85,7 @@ export class ChromeConnection {
     private static ATTACH_TIMEOUT = 10000; // ms
 
     private _socket: WebSocket;
+    private _crdpSocketMultiplexor: CRDPMultiplexor;
     private _client: Client;
     private _targetFilter: ITargetFilter;
     private _targetDiscoveryStrategy: ITargetDiscoveryStrategy;
@@ -100,23 +104,29 @@ export class ChromeConnection {
     /**
      * Attach the websocket to the first available tab in the chrome instance with the given remote debugging port number.
      */
-    public attach(address = '127.0.0.1', port = 9222, targetUrl?: string, timeout?: number): Promise<void> {
-        return this._attach(address, port, targetUrl, timeout)
+    public attach(address = '127.0.0.1', port = 9222, targetUrl?: string, timeout?: number, extraCRDPChannelPort?: number): Promise<void> {
+        return this._attach(address, port, targetUrl, timeout, extraCRDPChannelPort)
             .then(() => { });
     }
 
-    public attachToWebsocketUrl(wsUrl: string): void {
+    public attachToWebsocketUrl(wsUrl: string, extraCRDPChannelPort?: number): void {
         this._socket = new LoggingSocket(wsUrl);
-        this._client = new Client(<WebSocket>this._socket as any);
+        if (extraCRDPChannelPort) {
+            this._crdpSocketMultiplexor = new CRDPMultiplexor(this._socket);
+            new WebSocketToLikeSocketProxy(extraCRDPChannelPort, this._crdpSocketMultiplexor.addChannel('extraCRDPEndpoint')).start();
+            this._client = new Client(this._crdpSocketMultiplexor.addChannel('debugger'));
+        } else {
+            this._client = new Client(<WebSocket>this._socket as any);
+        }
 
         this._client.on('error', e => logger.error('Error handling message from target: ' + e.message));
     }
 
-    private _attach(address: string, port: number, targetUrl?: string, timeout = ChromeConnection.ATTACH_TIMEOUT): Promise<void> {
+    private _attach(address: string, port: number, targetUrl?: string, timeout = ChromeConnection.ATTACH_TIMEOUT, extraCRDPChannelPort?: number): Promise<void> {
         return utils.retryAsync(() => this._targetDiscoveryStrategy.getTarget(address, port, this._targetFilter, targetUrl), timeout, /*intervalDelay=*/200)
             .catch(err => Promise.reject(errors.runtimeConnectionTimeout(timeout, err.message)))
             .then(wsUrl => {
-                return this.attachToWebsocketUrl(wsUrl);
+                return this.attachToWebsocketUrl(wsUrl, extraCRDPChannelPort);
             });
     }
 
--- a/src/chrome/crdpMultiplexing/crdpMultiplexor.ts
+++ b/src/chrome/crdpMultiplexing/crdpMultiplexor.ts
@@ -5,6 +5,21 @@
 import { LikeSocket } from 'noice-json-rpc';
 import { logger } from 'vscode-debugadapter';
 
+/* Rational: The debug adapter only exposes the debugging capabilities of the Chrome Debugging Protocol.
+   We want to be able to connect other components to expose other of the capabilities to the users.
+   We are trying to do this by creating a multiplexor that will take one connection to a Chrome Debugging Protocol websocket,
+   and simulate from that that we have actually many independent connections to several Chrome Debugging Protocol websockets.
+   Given that implementing truly independent connections is not trivial, we are choosing to implement only the features that we
+   seem to need so far to support a component for an external Console, and an external DOM Explorer.
+
+   The way we are going to make that work is that we'll pass the actual websocket to chrome/node to the Multiplexor, and then we are going
+   to requests two channels from it. One will be the debugger channel, that will be used by the traditional debug adapter for everything it does.
+   Another channel will be the extraCRDPEndpoint channel that will be offered thorugh the client using a websocket port (so it'll simulate that
+   this is an actual connection to chrome or node), so we can connect any utilities that we want there.
+
+   In the future, if we need to, it's easy to modify this multiplexor to support more than 2 channels.
+*/
+
 /* Assumptions made to implement this multiplexor:
     1. The message IDs of CRDP don't need to be sent in order
     2. The Domain.enable messages don't have any side-effects (We might send them multiple times)
--- a/src/chrome/crdpMultiplexing/crdpMultiplexor.ts
+++ b/src/chrome/crdpMultiplexing/crdpMultiplexor.ts
@@ -35,7 +35,7 @@ function decodifyId(encodifiedId: number): number {
 }
 
 function throwCriticalError(message: string): void {
-    logger.log("CRDP Multiplexor - CRITICAL-ERROR: " + message);
+    logger.error("CRDP Multiplexor - CRITICAL-ERROR: " + message);
     throw new Error(message);
 }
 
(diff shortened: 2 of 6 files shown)


src/chrome/crdpMultiplexing/crdpMultiplexor.ts
```diff
@@ -0,0 +1,201 @@
...
+
+function decodifyChannelId(encodifiedId: number): number {
+    return encodifiedId % 10;
+}
+function decodifyId(encodifiedId: number): number {
+    return Math.floor(encodifiedId / 10);
+}
+
+function throwCriticalError(message: string): void {
+    logger.log("CRDP Multiplexor - CRITICAL-ERROR: " + message);
```
roblourens: `logger.error`?

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
```diff
@@ -0,0 +1,201 @@
...
+        // logger.log(`CRDP Multiplexor - Storing message to channel ${this.name} for ${domain} for later: ${messageData}`);
+        messagesForDomain.push(messageData);
+    }
+
+    constructor(public name: string, public id: number, private _multiplexor: CRDPMultiplexor) { }
+
+    public send(messageData: string): void {
+        const message = JSON.parse(messageData);
+        const method = message.method;
+        const isEnableMethod = method && method.indexOf(".enable") >= 0;
```
roblourens: Make sure it's distinguished from `".enableFoo"` just in case

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        this._multiplexor.send(this, messageData);
+
+        if (isEnableMethod) {
+            this.sendUnsentPendingMessages(domain);
+        }
+    }
+
+    private sendUnsentPendingMessages(domain: string): void {
```
roblourens: Is this to allow multiple channels to enable a particular domain? I'm not sure we should allow that, since you can't keep the state in sync between multiple things. Is there a use case right now?

src/chrome/chromeConnection.ts
```diff
@@ -100,23 +104,29 @@ export class ChromeConnection {
...
             .then(() => { });
     }
 
-    public attachToWebsocketUrl(wsUrl: string): void {
+    public attachToWebsocketUrl(wsUrl: string, extraCRDPChannelPort?: number): void {
         this._socket = new LoggingSocket(wsUrl);
-        this._client = new Client(<WebSocket>this._socket as any);
+        if (extraCRDPChannelPort) {
+            this._crdpSocketMultiplexor = new CRDPMultiplexor(this._socket);
+            new WebSocketToLikeSocketProxy(extraCRDPChannelPort, this._crdpSocketMultiplexor.addChannel('extraCRDPEndpoint')).start();
```
roblourens: Make sure this is shut down cleanly when debugging is done, even when running in debug server mode.

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
```diff
@@ -0,0 +1,201 @@
...
+    public on(event: "message", cb: (data: string) => void): void;
+    public on(event: string, cb: Function): void {
+        if (event === 'message') {
+            if (this._messageCallbacks.length === 0) {
+                setTimeout(() => this.discardUnsentPendingMessages(), CRDPChannel.timeToPreserveMessagesInMillis);
+            }
+
+            this._messageCallbacks.push(cb);
+        } else {
+            this._multiplexor.addListenerOfNonMultiplexedEvent(event, cb);
```
roblourens: Is this the right thing to do for `.on('open', ...`?

src/chrome/crdpMultiplexing/webSocketToLikeSocketProxy.ts
```diff
@@ -0,0 +1,32 @@
...
+    constructor(private _port: number, private _socket: LikeSocket) { }
+
+    public start(): void {
+        this._server = new WebSocket.Server({ port: this._port }, () => {
+            logger.log(`CRDP Proxy listening on: ${this._port}`);
+        });
+
+        this._server.on('connection', openedWebSocket => {
+            logger.log(`CRDP Proxy accepted a new connection`);
+            openedWebSocket.on('message', data => {
```
roblourens: I guess I'm not sure why you have this socket receiving messages and forwarding them to another socket, which is just open in the same process. ie, why is the Channel a LikeSocket?

src/chrome/crdpMultiplexing/webSocketToLikeSocketProxy.ts
```diff
@@ -0,0 +1,32 @@
...
+import * as WebSocket from 'ws';
+import { LikeSocket } from 'noice-json-rpc';
+
+export class WebSocketToLikeSocketProxy {
+    private _server: WebSocket.Server;
+
+    constructor(private _port: number, private _socket: LikeSocket) { }
+
+    public start(): void {
+        this._server = new WebSocket.Server({ port: this._port }, () => {
```
roblourens: Does this have to be a websocket? Does VS actually think it's connecting to Chrome, or does it just know it needs to connect on some port?

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
```diff
@@ -0,0 +1,201 @@
...
+    public on(event: string, cb: Function): void;
+    public on(event: "open", cb: (ws: LikeSocket) => void): void;
+    public on(event: "message", cb: (data: string) => void): void;
+    public on(event: string, cb: Function): void {
+        if (event === 'message') {
+            if (this._messageCallbacks.length === 0) {
+                setTimeout(() => this.discardUnsentPendingMessages(), CRDPChannel.timeToPreserveMessagesInMillis);
+            }
+
+            this._messageCallbacks.push(cb);
```
roblourens: When are there multiple message callbacks?

src/chrome/crdpMultiplexing/webSocketToLikeSocketProxy.ts
digeff: We didn't modify the existing logic at all, and the existing logic connects to a chrome-like websocket at a specified address e.g.: ws://localhost:1234/someUriHere so we need to maintain that unless we want to update the other side to support some different method of communication

src/chrome/crdpMultiplexing/webSocketToLikeSocketProxy.ts
digeff: The idea was to use the Multiplexor to put 1 websocket in, and get 2 websockets out, but given that the noice-json-rpc client uses LikeSockets and it seems like a simpler interface, I implemented the Multiplexor to put 1 LikeSocket in and get 2 (or n) LikeSockets out...
The reason we need the WebServer is because we need to open and listen in a port, to be able to communicate with the other process.

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
digeff: I wasn't sure if that was needed or not, and it seemed cheaper to just implement it rather than go and investigate if it's being used or not.

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
digeff: You are right, we cannot really sync the state between multiple things, but we are just choosing to ignore the bad side effects that problem brings from the time being. We currently need to enable "Runtime" for all the channels, so we do need to support the same Domain in different channels.

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
digeff: As you point out, this is not a good implementation for  .on('open', ...). If a channel is opened after a while it'll never get this callback fired. But it works for the time being for the way we are using the channels, so I'm inclined to leave this as-is for the time being unless you disagree...

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
digeff: Changed

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
digeff: Good catch, changing to endsWith...

src/chrome/chromeConnection.ts
digeff: I missed that... I'm going to add some logic to close all the websockets and the server when the Chrome WebSocket is closed...

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
roblourens: I guess you could, and here you can have more than one connection on the websocket server.

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
roblourens: Good point, we can't get around needing Runtime everywhere.

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
```diff
@@ -0,0 +1,201 @@
...
+        this._pendingMessagesForDomain = null;
+    }
+
+    public on(event: string, cb: Function): void;
+    public on(event: "open", cb: (ws: LikeSocket) => void): void;
+    public on(event: "message", cb: (data: string) => void): void;
+    public on(event: string, cb: Function): void {
+        if (event === 'message') {
+            if (this._messageCallbacks.length === 0) {
+                setTimeout(() => this.discardUnsentPendingMessages(), CRDPChannel.timeToPreserveMessagesInMillis);
```
roblourens: Do the DOM Explorer and Console connect separately, or share the same connection? Seems like you could lose these cached messages when there are two things behind a single Channel

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
```diff
@@ -0,0 +1,201 @@
...
+    if (methodParts.length === 2) {
+        return methodParts[0];
+    } else {
+        throwCriticalError(`The method ${method} didn't have exactly two parts`);
+        return "Unknown";
+    }
+}
+
+function encodifyChannel(channelId: number, id: number): number {
+    return id * 10 + channelId;
```
roblourens: I'm not sure about this, it would be easy to avoid. But I don't think Chrome will ever care whether message ids come in order. So, let's leave it.

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
digeff: They share the same connection, the same as the original implementation, that's why I'm assuming we won't run into issues there, but we'd need to test it to be sure...

src/chrome/crdpMultiplexing/crdpMultiplexor.ts
digeff: I like this approach because without any extra effort we can easily tell which channel sent each message...

src/chrome/crdpMultiplexing/webSocketToLikeSocketProxy.ts
```diff
@@ -0,0 +1,56 @@
...
+                    logger.log("CRDP Proxy succesfully shut down");
+                }
+
+                return {};
+            });
+        });
+
+        this._server.on('connection', openedWebSocket => {
+            if (this._currentlyOpenedWebSocket !== null) {
+                throw Error(`CRDP Proxy: Only one websocket is supported by the server on port ${this._port}`);
```
roblourens: You should close the socket here too

roblourens:
Overall I think this makes sense. I was a little confused at first about the Channels and why they are created statically, I was expecting something like that to be created on demand. But now I understand that and the separation between a channel and the websocket server that exposes it to VS and I think it all makes sense. It would be great if you can add a few comments in crdpMultiplexor.ts about the design and role of the multiplexor vs channel. And I'll wait to merge for your comment about closing everything down cleanly. Thanks!

src/chrome/crdpMultiplexing/webSocketToLikeSocketProxy.ts
digeff: Thanks, I missed that...

src/chrome/chromeConnection.ts
roblourens: Are you still going to do something for this?

Issue #215, opened 2017-07-07
Script Events should not send sourceReference for physical files


digeff:
Script Events should not send sourceReference for physical files to make sure that the intention is to read the file path rather than use the sourceReference.

dlebu:
This issue can be closed now.

PR #218, opened 2017-07-13
Script Events should not send sourceReference for physical files 

dlebu:
Fixing https://github.com/Microsoft/vscode-chrome-debug-core/issues/215
If the script url is a local path on the disk, we stop sending the sourceReference. @digeff @rakatyal

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -1365,21 +1365,19 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         const sourceReference = this.getSourceReferenceForScriptId(script.scriptId);
         const origin = this.getReadonlyOrigin(script.url);
 
-        return new Promise((resolve, reject) => {
-            utils.existsAsync(script.url)
-                .then((exists) => {
-                    const source = {
-                        name: path.basename(script.url),
-                        path: script.url,
-                        // if the path exists, do not send the sourceReference
-                        sourceReference: exists ? undefined : sourceReference,
-                        origin
-                    };
+        return utils.existsAsync(script.url)
+            .then((exists) => {
+                const source = {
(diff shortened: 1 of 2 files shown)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -1359,19 +1361,25 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
         const origin = this.getReadonlyOrigin(script.url);
-        const source = {
-            name: path.basename(script.url),
-            path: script.url,
-            sourceReference,
-            origin
-        };
 
-        const clientScript: Script = new Script(source);
+        return new Promise((resolve, reject) => {
```
roblourens: You don't need to create a new Promise, since you're working with `utils.existsAsync` that returns a promise.

src/utils.ts
```diff
@@ -36,6 +36,21 @@ export function existsSync(path: string): boolean {
...
+ * Checks asynchronously if a path exists on the disk.
+ */
+export function existsAsync(path: string): Promise<boolean> {
+    return new Promise((resolve, reject) => {
+        try {
+            fs.access(path, (err) => {
+                resolve(err ? false : true);
+            });
+        } catch (e) {
+            resolve(false);
```
digeff: Does access ever throw a sync exception?

rakatyal:
LGTM!

src/utils.ts
dlebu: The documentation doesn't mention it but I did not want to take any chances.

digeff: Maybe you can try sending null, an invalid path, or an {} as the path parameter to see if that throws an exception or returns an error in the callback...

dlebu: It throws for `null` and `undefined`.

Issue #205, opened 2017-05-31
"Could not download sourcemap from file:///..."


roblourens:
Seen in the tsc repo. This should be detected and loaded locally.

```
Debugging with inspector protocol because Node.js v8.0.0 was detected.
SourceMaps.loadSourceMapContents: Could not download sourcemap from file:////Users/roblou/code/TypeScript/built/local/cancellationToken.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from file:////Users/roblou/code/TypeScript/built/local/typescriptServices.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from file:////Users/roblou/code/TypeScript/built/local/tsserver.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from file:////Users/roblou/code/TypeScript/built/local/watchGuard.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from file:////Users/roblou/code/TypeScript/built/local/tsserverlibrary.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from file:////Users/roblou/code/TypeScript/built/local/typingsInstaller.js.map
SourceMaps.loadSourceMapContents: Could not download sourcemap from file:////Users/roblou/code/TypeScript/built/local/typescriptServices.js.map
```

PR #220, opened 2017-07-21
add support to eval code and solve file:// url bug

Alphapage:


Alphapage:
This PR could solve: 
https://github.com/Microsoft/vscode-chrome-debug-core/issues/205
https://github.com/Microsoft/vscode-node-debug2/issues/122
maybe https://github.com/Microsoft/vscode-chrome-debug/issues/460 and older issues I found.

Alphapage:
For info, I use global to register evaled sources. This is ugly, but modules were not working. Maybe different process !

Alphapage:
The main problems I was unable to solve:
- you need to  add `debugger;` at the beginning of eval code or add a breakpoint in a non eval code otherwise breakpoints will never be applied.
- it only works with node-debug2. node-debug or chrome-debug shows the evaled code instead of mapped one
- you need to show a link somewhere in console or in UI to be able to open the evaled code in a tab. I wasn't able to find a way to do it.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1423,6 +1423,10 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             // If the frame doesn't have a function name, it's either an anonymous function
             // or eval script. If its source has a name, it's probably an anonymous function.
             const frameName = functionName || (script.url ? '(anonymous function)' : '(eval code)');
+            if (script.hasSourceURL) {
+                if (!global["evalSources"]) global["evalSources"] = [];
```
roblourens: Please don't use `global`, try to find another way to share this list between modules.

src/sourceMaps/sourceMap.ts
```diff
@@ -161,6 +161,7 @@ export class SourceMap {
      * Will return null instead of a mapping on the next line (different from generatedPositionFor).
      */
     public authoredPositionFor(line: number, column: number): MappedPosition {
+        if (global['evalSources'] && global['evalSources'].find(s => this._generatedPath == s) && line > 0) line--; // check if eval code
```
roblourens: Why do you need to adjust the line differently based on whether it's an eval script or not?

src/sourceMaps/sourceMap.ts
```diff
@@ -223,7 +224,7 @@ export class SourceMap {
             return null;
         } else {
             return {
-                line: position.line - 1, // Back to 0-indexed lines
+                line: global["evalSources"] && global["evalSources"].find(s => s == this._generatedPath) ? position.line : position.line - 1, // check if eval code else Back to 0-indexed lines
```
roblourens: Same here - the 0-indexed or 1-indexed lines are a property of the source-map lib, it doesn't matter whether the code originated as an eval script or not.

roblourens:
Please see tslint issues as well

src/sourceMaps/sourceMap.ts
Alphapage: I think when code is an eval script, then `(function(){\n` is added and chrome debugger result is offset by 1 line vs source-map which use the script code.

src/chrome/chromeDebugAdapter.ts
Alphapage: I tried import/export without success. Do you have another idea ?

Alphapage:
I don't know how chrome debugger works, but the best way to solve this problem would be to send the eval file from //#sourceURL and sourcemap file from //#sourceMappingURL to chrome debugger and let chrome debugger do the job, apply breakpoints...
But I don't know if v8 debugger api provides such a feature.

src/sourceMaps/sourceMap.ts
Alphapage: Same as previous answer. I think either generated and authored code relay on`(function(){\n` problem.
in my opinion, it is more +1 indexed than 1-indexed: I keep the base index but add some adjustments.

PR #223, opened 2017-07-28
Added tests for the CRDP Multiplexor.

dlebu:
@digeff @roblourens

### DIFF
--- a/test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
+++ b/test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
@@ -1,0 +1,175 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { CRDPMultiplexor } from '../../../src/chrome/crdpMultiplexing/crdpMultiplexor';
+import { LikeSocket } from 'noice-json-rpc';
+import { Mock, Times, It } from 'typemoq';
+import * as Assert from 'assert';
+
+class StubSocket implements LikeSocket {
+
+    send(message: string): void { /* empty */ }
+    removeListener(event: string, cb: Function): any { /* empty */ }
+    on(event: 'open', cb: (ws: LikeSocket) => void): any;
+    on(event: 'message', cb: (data: string) => void): any;
+    on(event: string, cb: Function): any { /* empty */ }
+}
+
+suite('CRDPMultiplexor', () => {
+
+    let multiplexor: CRDPMultiplexor;
+    let webSocketMock: Mock<StubSocket>;
+    let socketMessageCallbacks: Function[];
+
+
+    setup(() => {
+        let socket = new StubSocket();
+        socketMessageCallbacks = [];
+        webSocketMock = Mock.ofInstance<StubSocket>(socket);
+        webSocketMock
+            .setup(s => s.on('message', It.isAny()))
+            .returns((s: string, cb: any) => {
+                socketMessageCallbacks.push(cb);
+            });
+        webSocketMock
+            .setup(s => s.send(It.isAnyString()))
+            .returns((s: string) => {
+                let message = JSON.parse(s);
+                for (let cb of socketMessageCallbacks) {
+                    // Send empty result once we received a message
+                    cb('{"id":' + message.id + ', "result":{}}');
+                }
+            });
+        multiplexor = new CRDPMultiplexor(webSocketMock.object);
+    });
+
+    teardown(() => {
+        multiplexor = undefined;
+        webSocketMock.reset();
+        webSocketMock = undefined;
+        socketMessageCallbacks = [];
+    });
+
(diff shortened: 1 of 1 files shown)


test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+
+    setup(() => {
+        let socket = new StubSocket();
+        socketMessageCallbacks = [];
+        webSocketMock = Mock.ofInstance<StubSocket>(socket);
+        webSocketMock
+            .setup(s => s.on('message', It.isAny()))
+            .returns((s: string, cb: any) => {
+                socketMessageCallbacks.push(cb);
```
digeff: Why not implement this directly inside of StubSocket  instead of using a Mock? You can store all the received or sent messages in an array if you want to verify them

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
```diff
@@ -0,0 +1,175 @@
...
+                let callback = (data: string) => {
+                    Assert.equal(JSON.parse(data).id, i);
+                    resolve();
+                }
+
+                channel.on("message", callback);
+            });
+
+            callbackPromises.push(callbackPromise);
+            multiplexor.send(channel, '{"method":"Console.enable","id":' + i + '}');
```
digeff: It would also be interesting to test that if all channels use the same message id, things still work

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
```diff
@@ -0,0 +1,175 @@
...
+                        Assert.equal(notification.method, testNotification);
+                        resolve();
+                    }
+                }
+
+                channel.on("message", callback);
+            });
+
+            // Enable the Test domain so we will receive notifications
+            channel.send('{"method":"' + testEnable + '","id":' + i + '}');
```
digeff: Even though we are enabling the domains in sync way now, we might convert it to async in the future. It'd be great if the same test still worked...

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
```diff
@@ -0,0 +1,175 @@
...
+        let domain1Enable = "Domain1.enable";
+        let domain2Enable = "Domain2.enable";
+        let domain1Notification = "Domain1.notification";
+        let domain2Notification = "Domain2.notification";
+
+        let domain2Enabled = false;
+
+        let channel = multiplexor.addChannel("channel");
+        channel.on("message", (data: string) => {
+            let notification = JSON.parse(data);
```
digeff: It's difficult for me to follow this logic... I think it might be easier for me to follow if we logged all the messages in an array, and then we just verified the contents at the end...

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
dlebu: I wanted to keep the StubSocket class empty. I only created it because `typemoq` wouldn't allow me to mock an interface. So think of it as the real `Socket` class.

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
dlebu: That would be easy to promisify when we change the API.

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
dlebu: Adding a new test for that.

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
digeff: :thumbsup:

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
digeff: :thumbsup:

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
dlebu: I changed it with an array. Good idea.

test/chrome/crdpMultiplexing/crdpMultiplexor.test.ts
```diff
@@ -139,24 +151,16 @@ suite('CRDPMultiplexor', () => {
         let domain2Notification = "Domain2.notification";
 
         let domain2Enabled = false;
+        let receivedMessages = [];
```
digeff: This is much easier to understand :)

digeff:
👍

roblourens:
Could you check out the tslint issues in Travis?

dlebu:
@roblourens Done, thanks!

Issue #226, opened 2017-08-09
'Debugger.setAsyncCallStackDepth' wasn't found


iTOYS:
`Debugger.setAsyncCallStackDepth` method is called during attach and while it is [marked as optional](https://github.com/Microsoft/vscode-chrome-debug-core/blob/012a56f8637a380c2117b6fecd096e5412925d08/crdp/crdp.d.ts#L1308) its existence is [not being checked](https://github.com/Microsoft/vscode-chrome-debug-core/blob/50b5fffe4ebc3edb7bf87cc483eb6ec758d361d5/src/chrome/chromeDebugAdapter.ts#L342) before calling.

iTOYS:
#https://github.com/Microsoft/vscode-cordova/issues/297

PR #227, opened 2017-08-09
Check if 'Debugger.setAsyncCallStackDepth' method unimplemented

iTOYS:
Fix for #226


roblourens:
This won't fix it, the problem is that the runtime doesn't implement it. You need to call the method and ignore the error if it returns failure.

iTOYS:
I already fixed it in my code. Just ignore that error. But I thought it's correct to check if method implemented.

roblourens:
That check doesn't tell you if it's implemented, that object is just a proxy calling the target with whatever you give it.

iTOYS:
Oh, I see. It won't work(

PR #228, opened 2017-08-11
Set `supportsValueFormatting` flag in initialize

andrewcrawley:
Support for stack trace formatting was added a while ago, but the Capabilities flags weren't updated to reflect this.  The VS Debug Adapter Host is being updated to only send formatting data to adapters that indicate support.


roblourens:
Thanks!

PR #238, opened 2017-09-15
Now we unload scripts when we change to a different page

digeff:
* Replaced ScriptEvent method with new LoadedSourceEvent
* Now we also send the removed events


digeff:
@rakatyal , @roblourens : Could you please take a look?

PR #246, opened 2017-10-05
Fix site.js appearing in the call stack instead of site.ts

digeff:
The code wasn't awaiting, so part of the code to create the response run before resolving the source maps, and the other part after it so we ended up with part saying site.js and part saying site.ts


roblourens:
Good catch!

digeff:
@roblourens Is there any option for tslint that can catch these kind of things?

roblourens:
Not every async function should be awaited, it's tough to know just from looking at the code whether it's right or wrong.

digeff:
In C# you call .Forget() to signal that you are "not awaiting it" on purpose...

PR #247, opened 2017-10-12
Added a new FallbackToClientPathTransformer that will use the UrlPathTransformer and fallback to the client if neccesary

digeff:
For some programs the debugger might not have enough information to figure out how to map all the URLs to file paths, so we can ask the client if he knows how to do it

@roblourens @rakatyal

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -28,6 +28,7 @@ import {BasePathTransformer} from '../transformers/basePathTransformer';
 import {RemotePathTransformer} from '../transformers/remotePathTransformer';
 import {BaseSourceMapTransformer} from '../transformers/baseSourceMapTransformer';
 import {EagerSourceMapTransformer} from '../transformers/eagerSourceMapTransformer';
+import {FallbackToClientPathTransformer} from '../transformers/fallbackToClientPathTransformer';
 
 import * as path from 'path';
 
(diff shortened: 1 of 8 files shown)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -220,6 +221,10 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             });
         }
 
+        if (args.supportsMappingURLsToFilePaths) {
+            this._pathTransformer = new FallbackToClientPathTransformer(this._session);
```
roblourens: Do this at the top of the method so we are less likely to add some code working on pathTransformer before this.

src/transformers/fallbackToClientPathTransformer.ts
```diff
@@ -0,0 +1,55 @@
...
+                        resolve(err ? false : true);
+                    });
+                } catch (e) {
+                    resolve(false);
+                }
+            }).then(doesFilePathExist => doesFilePathExist
+                // If it does, we use that result
+                ? filePath
+                // If it doesn't, we ask the client to map it as a fallback
+                : this.requestClientToMapURLToFilePath(scriptUrl).catch(rejection => {
```
roblourens: Can include eval scripts which are given a fake url like `VM45`. Not sure whether you want to send those back to the client.

digeff: That is a good question. I'll consult with a PM.

src/chrome/chromeUtils.ts
```diff
@@ -281,3 +281,9 @@ export function selectBreakpointLocation(lineNumber: number, columnNumber: numbe
 
     return locations[0];
 }
+
+export var EVAL_NAME_PREFIX = 'VM';
```
roblourens: const

PR #249, opened 2017-10-15
Suppress extraneous newline in console output

jpap:
When using this debug adapter with Chrome/Electron and [winston-electron](https://github.com/dustinblackman/winston-electron), I get extraneous newlines throughout the console.

This PR removes those extraneous newlines introduced by this debug adapter.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -902,7 +902,8 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
                 msg = await this.mapFormattedException(msg);
             }
 
-            e = new OutputEvent(msg + '\n', category);
+            const optionalNewLine = !msg.endsWith('\n') ? '\n' : '';
+            e = new OutputEvent(msg + optionalNewLine, category);
         } else {
             e = new OutputEvent('output', category);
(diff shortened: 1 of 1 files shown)


roblourens:
Why should we have to do that? If someone wants to log an extra newline, they should be able to do that. This would make `console.log('foo')` the same as `console.log('foo\n')`. Where are the extra newlines actually coming from?

jpap:
Agree with your comment however I think they're coming from here -- as they get inserted here. I don't see the same newlines in the DevTools debugger in Chrome itself.

jpap:
... actually, hang tight; I need to investigate a bit more. It looks like the following do not introduce the newlines, so it might be coming from Winston:

```js
    console.log("test: log");
    console.debug("test: debug");
    console.error("test: error");
```

roblourens:
If you can provide a code sample with winston that produces an unexpected result (or a different result from chrome devtools) I'd like to try it out.

jpap:
My apologies; I [found the culprit in winston-electron](https://github.com/dustinblackman/winston-electron/blob/0c16ad190fd52f4f59a94244d1ab9c76042bbd35/index.js#L156) and missed it in an earlier search because I was looking for `\n` and not `this.eol` (`os.EOL`).

jpap:
For reference: [winston-electron#4](https://github.com/dustinblackman/winston-electron/issues/4)

Issue #256, opened 2017-11-15
Cannot read property 'mappedPath' of undefined  at EagerSourceMapTransformer.setBreakpoints


auchenberg:
From Mike Kaufman

>Hey Kenneth – just installed new version of VS code yesterday and now I can’t set breakpoints.  I see this in my output:
 
```
Debugging with inspector protocol because Node.js v8.6.0 was detected.
node --inspect-brk=40654 src\agent.js
Debugger listening on ws://127.0.0.1:40654/9fe6fed3-1710-4ef6-859f-f9f26d230ecb
Error processing "setBreakpoints": TypeError: Cannot read property 'mappedPath' of undefined
    at EagerSourceMapTransformer.setBreakpoints (c:\Program Files\Microsoft VS Code\resources\app\extensions\ms-vscode.node-debug2\node_modules\vscode-chrome-debug-core\out\src\transformers\baseSourceMapTransformer.js:63:23)
    at validateBreakpointsPath.then (c:\Program Files\Microsoft VS Code\resources\app\extensions\ms-vscode.node-debug2\node_modules\vscode-chrome-debug-core\out\src\chrome\chromeDebugAdapter.js:827:40)
    at <anonymous>
```

Known issue?  We’re generating maps w/ gulp-typescript IIRC, so maybe something w/ how maps are getting gen’d?

roblourens:
Can I have the log or repro steps?

auchenberg:
Ping @mike-kaufman

mike-kaufman:
how would I create a log?  Repo steps may involve a private repo, so I can PM you instructions for that if you'd like.

roblourens:
You can just set `"trace": true` in your launch config, then reproduce the issue, then upload the log from the path it prints in the debug console. I would take repro steps too over email, whichever is easier.

roblourens:
The bug is

- Have a BP set in a doc that the adapter hasn't mapped to disk
- Stop debugging, keep the doc open
- Start debugging again

The doc's sourceReference is from the previous debug session. The adapter isn't expecting a sourceReference that doesn't exist yet.

I will reply over email about why the sourcemaps aren't working.

roblourens:
Error messages will be fixed in the next release, but the underlying issue could still be addressed.

tgshub:
I am getting the similar error, i guess its because of the change in the node version, I had node v8.0 now its v10.x
Debugging with inspector protocol because Node.js v10.16.0 was detected.

PR #263, opened 2018-01-11
fix: handle is undefined

obastemur:
Debugging on macOS with node 8.7, receiving the error that is originating from `handle is undefined`. I'm not sure about the reasons but simple check fixes the debug support for me.


roblourens:
Probably https://github.com/Microsoft/vscode-chrome-debug-core/issues/256#issuecomment-345868276. I would rather fix it a different way, but this is fine to get rid of the error message for now.

roblourens:
It was merged more than 2 years ago. Please file an issue with more details about whatever issue you are having.

TCourtneyOwen:
@roblourens sorry I noticed that a few minutes after I submitted the comment and have updated my package version. Thanks!

PR #264, opened 2018-01-12
Don't add the same breakpoint id to cache.

changsi-an:
Or otherwise the same breakpoint id will be registered twice. This can happen when EDP choose to send breakpointResolved event even if the response for setBreakpointByUrl already contains the information for the same bound breakpoint.


roblourens:
> This can happen when EDP choose to send breakpointResolved event even if the response for setBreakpointByUrl already contains the information for the same bound breakpoint.

Is this a bug with EDP?

changsi-an:
Yes, it's a known issue on EDP side. There is a bug tracking that.

PR #265, opened 2018-01-12
fix undefined variable exception

obastemur:
Exception appears after editing breakpoints list from the editor. I don't have a concrete way to repro but submitting this fix after having the same problem for the third time. Simply, `handle` is `undefined` hence reaching out `scriptId` throws an exception.


roblourens:
Thanks for the PR! Also https://github.com/Microsoft/vscode-chrome-debug-core/issues/256

PR #267, opened 2018-01-18
Expose the connection object, so the derivative class can revisit the…

changsi-an:
Expose the connection object, so the derivative class can revisit the targets as needed.

Currently the series of attach() method could not surface back the debugger id from the response of calling /json/list. Instead of refactoring the code and returning that, it might be cleaner to have the downstream logic query the targets again when needed.

### DIFF
--- a/src/chrome/chromeConnection.ts
+++ b/src/chrome/chromeConnection.ts
@@ -30,7 +30,7 @@ export interface ITarget {
 
 export type ITargetFilter = (target: ITarget) => boolean;
 export interface ITargetDiscoveryStrategy {
-    getTarget(address: string, port: number, targetFilter?: ITargetFilter, targetUrl?: string): Promise<string>;
+    getTarget(address: string, port: number, targetFilter?: ITargetFilter, targetUrl?: string): Promise<ITarget>;
     getAllTargets(address: string, port: number, targetFilter?: ITargetFilter, targetUrl?: string): Promise<ITarget[]>;
 }
 
@@ -94,6 +94,7 @@ export class ChromeConnection {
     private _client: Client;
     private _targetFilter: ITargetFilter;
     private _targetDiscoveryStrategy: ITargetDiscoveryStrategy;
+    private _attachedTarget: ITarget;
 
     constructor(targetDiscovery?: ITargetDiscoveryStrategy, targetFilter?: ITargetFilter) {
         this._targetFilter = targetFilter;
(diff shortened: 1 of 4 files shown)


roblourens:
Why do you need this? Do you need the attached target id, not the list of all target ids?

changsi-an:
"Do you need the attached target id" 

Yes

roblourens:
So this won't work, because it will return all targets. You need to change the attach logic to make the selected target accessible. What's the use case?

changsi-an:
This PR only reflects the changes needed in chrome-core. My debug adapter will take care of the attach logic. It will call `chromeConnection.attachedWebSocketUrl` to get the current attached target wsUrl. then match that from the result of getting all targets. I didn't want to infuse this into chrome-core, because such need is specific to my debug adapter. The reason I do this PR this way hoping to make it open and general, so it can be useful for other purposes instead of scoped to a limited use case.

The code in my downstream debug adapter looks like this:

```

               this.doAttach(args...)
               .then(() => {
                    if (!this._chromeConnection.isAttached) {
                        throw coreUtils.errP(localize("edge.debug.error.notattached", "Debugging connection is not attached after the attaching process."));
                    }

                    let websocketUrl = this._chromeConnection.attachedWebSocketUrl;
                    this._debugProxyPort = port;

                    return this._chromeConnection.getAllTargets(args.address, port, (target: chromeConnection.ITarget) => {
                        return target.webSocketDebuggerUrl == websocketUrl;
                    })
                }).then((targets: chromeConnection.ITarget[]) => {
                    if (targets.length != 1) {
                        throw coreUtils.errP(localize("edge.debug.error.noSoleMatchingTarget", "Cannot determinte the attached target."));
                    }

                    this._debuggerId = targets[0].id;
                });
```

roblourens:
I can't think of any other reason for needing the websocket url anyway. Maybe exposing the whole attached target object would be useful. But please tell me what your actual use case is, because I don't understand why you need this id.

changsi-an:
This id is used to call a API provided in EDP to close a tab. The API has the form of *:9222/json/close/{id} . Hence the id is used there. We can't just breakdown Edge process entirely like we do for Chrome due to different process model is used for each browser. 

I feel that dynamically retrieving the target is better than using the cached one. It's going to be a long time between the target is attached and when we need to use this id (debug finish). So far I am not sure whether edge is going to change any metadata of the target along the way. Even if the id might be fixated, I am not sure about the other metadata we might need to access along the way.

roblourens:
Good point that it will change along the way. I thought of one thing that would actually be kind of cool - we could show the window title in the call stack (while debugging multiple targets) instead of just "Chrome". Anyway, could you just change it to expose the attached ITarget?

changsi-an:
Exposing as ITarget is better. Will do!

changsi-an:
PTAL

PR #269, opened 2018-01-19
Make reporting loadedScource event customizable.

changsi-an:
So the subclass can override the implementation of `sendLoadedSourceEvent()` and decide what the reason is.


auchenberg:
What's the scenario for this?

PR #271, opened 2018-01-22
Add a unit test for the overriding behavior of sendLoadedSourceEvent()

changsi-an:
Added 1 more test.

### DIFF
--- a/test/chrome/chromeDebugAdapter.test.ts
+++ b/test/chrome/chromeDebugAdapter.test.ts
@@ -24,7 +24,7 @@ import * as testUtils from '../testUtils';
 import * as utils from '../../src/utils';
 
 /** Not mocked - use for type only */
-import {ChromeDebugAdapter as _ChromeDebugAdapter} from '../../src/chrome/chromeDebugAdapter';
+import {ChromeDebugAdapter as _ChromeDebugAdapter, LoadedSourceEventReason} from '../../src/chrome/chromeDebugAdapter';
 
 const MODULE_UNDER_TEST = '../../src/chrome/chromeDebugAdapter';
 suite('ChromeDebugAdapter', () => {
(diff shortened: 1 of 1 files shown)


changsi-an:
@digeff PTAL

test/chrome/chromeDebugAdapter.test.ts
```diff
@@ -427,6 +427,30 @@ suite('ChromeDebugAdapter', () => {
             });
         });
 
+        // This is needed for Edge debug adapter, please keep the signature and logic of sendLoadedSourceEvent() method intact.
+        test('tests that sendLoadedSourceEvent can accept an additional paramter to override the `reason` parameter', done => {
+            const loadedSourceEventReason: LoadedSourceEventReason = 'changed';
+            sendEventHandler = (event) => {
+                assert.equal('loadedSource', event.event);
+                assert.notEqual(null, event.body);
```
digeff: Would the test catch more potential issues if we tested the contents of all the members of the event argument?

test/chrome/chromeDebugAdapter.test.ts
```diff
@@ -427,6 +427,30 @@ suite('ChromeDebugAdapter', () => {
...
+                assert.equal(loadedSourceEventReason, event.body.reason);
+                done();
```
digeff: How will done be called if the test fails?

changsi-an: It will timeout and show as a timeout failure.

test/chrome/chromeDebugAdapter.test.ts
changsi-an: Should we keep the tests atomic and follow the one-thing-at-a-time principle? Or otherwise, when the product code changes, we might have to update all the tests, which is tedious and does not gain us anything.

digeff: Ideally the test should only fail when the contract/interface changes. If we change the implementation while maintaining the original contract/interface then the test shouldn't fail nor need updating.

test/chrome/chromeDebugAdapter.test.ts
digeff: It'd be great if we could show the actual failure rather than a timeout.

test/chrome/chromeDebugAdapter.test.ts
changsi-an: Resolved offline. @digeff  and I both agree on this change now.

test/chrome/chromeDebugAdapter.test.ts
changsi-an: Updated to return the promise directly instead of using `done` method.

PR #274, opened 2018-01-29
Block while parsing scripts to prevent sending 'removes' before 'adds'

digeff:
We were running into an issue where we send 'remove' myscript.js before we sent the 'add' myscript.js when we went into a page, and quickly redirected into another page.

That happened because we have some async code in how we process onScriptParsed, so that processing can be on-flight, and onExecutionContextsCleared very quickly sends the 'remove' events, and we weren't waiting for the onScriptParsed events to be completed.
This fixes solve part of the issue.

I think there is another unrelated reason that I'm investigating after this.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -459,11 +459,11 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             const initialSourceMapsP = this._initialSourceMapsP;
             this._initialSourceMapsP = null;
 
-            await initialSourceMapsP.then(async () => {
-                this._session.sendEvent(new InitializedEvent());
-                await Promise.all(this._earlyScripts.map(async script => await this.sendLoadedSourceEvent(script)));
-                this._earlyScripts = null;
-            });
+            await initialSourceMapsP;
+
+            this._session.sendEvent(new InitializedEvent());
+            await Promise.all(this._earlyScripts.map(script => this.sendLoadedSourceEvent(script)));
+            this._earlyScripts = null;
         }
     }
 
@@ -475,17 +475,19 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
      * e.g. the target navigated
      */
     private onExecutionContextsCleared(): Promise<void> {
-            const asyncOperations = [];
-            this._scriptsById.forEach(scriptedParseEvent => {
-                return this.doAfterProcessingSourceEvents(async () => { // This will not execute until all the on-flight 'new' source events have been processed
-                    asyncOperations.push(this.scriptToLoadedSourceEvent('removed', scriptedParseEvent).then(scriptEvent => {
-                        this._session.sendEvent(scriptEvent);
-                    }));
-                });
+        const asyncOperations = [];
+        this._scriptsById.forEach(scriptedParseEvent => {
+            return this.doAfterProcessingSourceEvents(async () => { // This will not execute until all the on-flight 'new' source events have been processed
+                asyncOperations.push(this.scriptToLoadedSourceEvent('removed', scriptedParseEvent).then(scriptEvent => {
+                    this._session.sendEvent(scriptEvent);
+                }));
             });
+        });
 
+        return this.doAfterProcessingSourceEvents(async () => { // Execute after sending all the 'removed' events
             return Promise.all(asyncOperations).then(() =>
                 this.clearTargetContext());
+        });
     }
 
     protected async onPaused(notification: Crdp.Debugger.PausedEvent, expectingStopReason = this._expectingStopReason): Promise<void> {
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -136,9 +136,9 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
     private _breakOnLoadHelper: BreakOnLoadHelper | null;
 
-    // Queue to synchronize onScriptParsed events and onExecutionContextsCleared events so that 'remove' script events
+    // Queue to synchronize new source loaded and source removed events so that 'remove' script events
     // won't be send before the corresponding 'new' event has been sent
-    private _scriptParsedQueue: Promise<any> = Promise.resolve(null);
+    private _sourceLoadedQueue: Promise<void> = Promise.resolve(null);
 
     public constructor({ chromeConnection, lineColTransformer, sourceMapTransformer, pathTransformer, targetFilter, enableSourceMapCaching }: IChromeDebugAdapterOpts, session: ChromeDebugSession) {
         telemetry.setupEventHandler(e => session.sendEvent(e));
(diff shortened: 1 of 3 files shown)


roblourens:
I don't think this will fix your problem. `await` won't keep the code in `onExecutionContextsCleared` from running. Your change will keep the rest of the code in onScriptParsed from running before all the scriptParsed notifications are sent, but won't stop other event handlers. 

You probably want a promise that you use as an async queue, like

```
private _scriptParsedQueue: Promise<any>;

...

onScriptParsed() {
  this._scriptParsedQueue = this._scriptParsedQueue.then(() => send load events)
}

onExecutionContextsCleared() {
  this._scriptParsedQueue.then(() => send delete events)
}

```

roblourens:
The problem is that we are getting a `breakpointResolved` event for the new script before you have processed the scriptLoaded event, so the breakpointResolved handler stops at this line: https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/chrome/chromeDebugAdapter.ts#L930

Two options:
1. Only delay the sending of the events to the client, not the whole onScriptParsed handler
1. Have the breakpointResolved handler also wait on the scriptParsedQueue

I think the first one is better for perf, and would hopefully prevent other issues with things that depend on the _scriptsById map.

roblourens:
Could you add a test to the PR as well, that adds some scripts, refreshes the page, and ensures that the script load/unload events are sent in the proper order? You can take some inspiration from this failing test on how to do that.

digeff:
Ran into a bug with the newest typescript: https://github.com/Microsoft/TypeScript/issues/21491
changing to an older version

src/chrome/chromeDebugAdapter.ts
```diff
@@ -449,32 +453,39 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
      */
-    protected sendInitializedEvent(): void {
+    protected async sendInitializedEvent(): Promise<void> {
         // Wait to finish loading sourcemaps from the initial scriptParsed events
         if (this._initialSourceMapsP) {
             const initialSourceMapsP = this._initialSourceMapsP;
             this._initialSourceMapsP = null;
 
-            initialSourceMapsP.then(() => {
+            await initialSourceMapsP.then(async () => {
```
roblourens: The mixing of async/await and promises is a little confusing here (`await ... .then(`)

src/chrome/chromeDebugAdapter.ts
```diff
@@ -449,32 +453,39 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
                 this._session.sendEvent(new InitializedEvent());
-                this._earlyScripts.forEach(script => this.sendLoadedSourceEvent(script));
+                await Promise.all(this._earlyScripts.map(async script => await this.sendLoadedSourceEvent(script)));
```
roblourens: The async/await inside `map` is redundant, just return the promise from this.sendLoadedSourceEvent

src/chrome/chromeDebugAdapter.ts
```diff
@@ -449,32 +453,39 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
                 this._earlyScripts = null;
             });
         }
     }
 
+    public doAfterProcessingSourceEvents(action: () => void): Promise<void> {
```
roblourens: `action` returns a Promise

src/chrome/chromeDebugAdapter.ts
```diff
@@ -449,32 +453,39 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
+                return this.doAfterProcessingSourceEvents(async () => { // This will not execute until all the on-flight 'new' source events have been processed
+                    asyncOperations.push(this.scriptToLoadedSourceEvent('removed', scriptedParseEvent).then(scriptEvent => {
+                        this._session.sendEvent(scriptEvent);
+                    }));
+                });
+            });
 
-        Promise.all(asyncOperations).then(() =>
-            this.clearTargetContext());
+            return Promise.all(asyncOperations).then(() =>
```
roblourens: This is confusing, `asyncOperations` will not be populated by this point because it's populated by an async function that we don't wait on.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -449,32 +453,39 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
      * e.g. the target navigated
      */
-    private onExecutionContextsCleared(): void {
-        const asyncOperations = [];
-        this._scriptsById.forEach(scriptedParseEvent => {
-            asyncOperations.push(this.scriptToLoadedSourceEvent('removed', scriptedParseEvent).then(scriptEvent =>
-                this._session.sendEvent(scriptEvent)));
-        });
+    private onExecutionContextsCleared(): Promise<void> {
+            const asyncOperations = [];
```
roblourens: I think this method picked up an extra level of indentation

src/chrome/chromeDebugAdapter.ts
digeff: Done

src/chrome/chromeDebugAdapter.ts
digeff: Done

src/chrome/chromeDebugAdapter.ts
digeff: The use inside onExecutionContextsCleared() doesn't return a promise

src/chrome/chromeDebugAdapter.ts
digeff: Done

src/chrome/chromeDebugAdapter.ts
digeff: Nice catch... I missed this...

src/chrome/chromeDebugAdapter.ts
```diff
@@ -449,32 +453,41 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
+            return this.doAfterProcessingSourceEvents(async () => { // This will not execute until all the on-flight 'new' source events have been processed
+                asyncOperations.push(this.scriptToLoadedSourceEvent('removed', scriptedParseEvent).then(scriptEvent => {
+                    this._session.sendEvent(scriptEvent);
+                }));
+            });
         });
 
-        Promise.all(asyncOperations).then(() =>
-            this.clearTargetContext());
+        return this.doAfterProcessingSourceEvents(async () => { // Execute after sending all the 'removed' events
```
roblourens: This is a confusing flow now. The asyncOperations array isn't needed if doAfterProcessingSourceEvents is forcing this to happen after sending the script removed events are sent.

Also, you could iterate the scripts inside the doAfterProcessingSourceEvent. That will be faster because the script processing doesn't have to happen one by one. You can use the `await Promise.all(scriptById.map(async () => ...` pattern again.

digeff: I rewrote this to use async/wait. What do you think?

PR #277, opened 2018-02-05
Make sure to return 1 response for each breakpoint

digeff:
Temporary fix for this issue.

Some clients fail if we don't return 1 response per request

### DIFF
--- a/src/chrome/breakOnLoadHelper.ts
+++ b/src/chrome/breakOnLoadHelper.ts
@@ -2,6 +2,7 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
+import {DebugProtocol} from 'vscode-debugprotocol';
 import {logger} from 'vscode-debugadapter';
 import {ISetBreakpointResult, BreakOnLoadStrategy} from '../debugAdapterInterfaces';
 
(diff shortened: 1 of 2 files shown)


roblourens:
Is this blocking or can we come up with a better solution?

rakatyal:
@roblourens: It is blocking our tests inside VS because the debugger errors out and exits if we send bak one response for multiple breakpoints inside the setBreakpoint request. 
I think this is not a neat way to do this. We should probably come up with some internal breakpoint structure which we can return for these requests. Would you know if something like this exists in the protocol?
But to unblock our tests, I am fine with this fix temporarily until we come up with a better way to do it. Thoughts?

roblourens:
Why weren't the tests failing before?

We need to send unverified responses for the other breakpoints. There are a couple helpers for this in ChromeDebugAdapter. That might be happening with this fix automatically.

roblourens:
I think that pushing empty objects is probably the right solution. What should happen, is we hit this line later: https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/chrome/chromeDebugAdapter.ts#L1358 and return unverified BPs for each.

roblourens:
Would be better if it's in the breakOnLoadHelper though

roblourens:
Oh and also it should not be returning the actual response from the break on load BP. That could make the first BP look like it was moved to line 1.

digeff:
Done

rakatyal:
Perfect!

roblourens:
Actually you should still await `addStopOnEntryBreakpoint`, right? I can add that.

roblourens:
Just published -core version 3.22.5

digeff:
Please add it, I missed that..

roblourens:
I did, it's in that release.

digeff:
Thanks :)

PR #278, opened 2018-02-05
Allows the child debug adapter to override some variable handling logic.

changsi-an:


roblourens:
Please, with all these PRs, explain what you're trying to do and why. I'm not going to accept PRs that make a bunch of random changes without explaining why they are necessary.

changsi-an:
Sorry, it's sent out by mistake.

PR #282, opened 2018-02-09
Fix breakpoint not hitting in refresh because it's being set on a negative line

digeff:
Breakpoints weren't being hit after refresh because BreakOnLoadHelper.resolvePendingBreakpoints was passing a value from inside this._chromeDebugAdapter.pendingBreakpointsByUrl to chromeDebugAdapter.setBreakpoints, and setBreakpoints was calling LineColTransformer and substracting 1 to the line number again.

The deep copy on args prevents setBreakpoints from modifying the original args...


roblourens:
What's the problem exactly and why does this fix it? It sounds like you're describing the problem that this was originally added to fix, but I can't repro any issue now.

PR #285, opened 2018-02-14
Fix break-on-load breakpoint not hitting on the first line for Chrome

digeff:
The previous userBreakpointOnLine1Col1 logic depended on parts of the ScriptParsed method executing synchronically to set that instance variable before ScriptPaused started executing. In recent changes, I added some new awaits to ScriptParsed which make it not execute synchronical any more, so the existing logic doesn't work any more.

### DIFF
--- a/src/chrome/breakOnLoadHelper.ts
+++ b/src/chrome/breakOnLoadHelper.ts
@@ -66,7 +66,7 @@ export class BreakOnLoadHelper {
         }
     }
 
-    private getPausedScriptUrlFromId(scriptId: string): string {
+    private getScriptUrlFromId(scriptId: string): string {
         return this._chromeDebugAdapter.scriptsById.get(scriptId).url;
     }
 
@@ -75,7 +75,7 @@ export class BreakOnLoadHelper {
      * Used when break on load active, either through Chrome's Instrumentation Breakpoint API or the regex approach
      */
     private async resolvePendingBreakpointsOfPausedScript(scriptId: string): Promise<boolean> {
-        const pausedScriptUrl = this.getPausedScriptUrlFromId(scriptId);
+        const pausedScriptUrl = this.getScriptUrlFromId(scriptId);
         const sourceMapUrl = this._chromeDebugAdapter.scriptsById.get(scriptId).sourceMapURL;
         const mappedUrl = await this._chromeDebugAdapter.pathTransformer.scriptParsed(pausedScriptUrl);
         let breakpointsResolved = false;
@@ -133,7 +133,7 @@ export class BreakOnLoadHelper {
         // using committedBreakpointsByUrl for our logic.
         let anyPendingBreakpointsResolved = await this.resolvePendingBreakpointsOfPausedScript(pausedLocation.scriptId);
 
-        const pausedScriptUrl = this.getPausedScriptUrlFromId(pausedLocation.scriptId);
+        const pausedScriptUrl = this.getScriptUrlFromId(pausedLocation.scriptId);
         // Important: We need to get the committed breakpoints only after all the pending breakpoints for this file have been resolved. If not this logic won't work
         const committedBps = this._chromeDebugAdapter.committedBreakpointsByUrl.get(pausedScriptUrl);
         const anyBreakpointsAtPausedLocation = committedBps.filter(bp =>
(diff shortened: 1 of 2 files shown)


src/chrome/breakOnLoadHelper.ts
```diff
@@ -122,16 +125,24 @@ export class BreakOnLoadHelper {
...
-        let anyPendingBreakpointsResolved = await this.resolvePendingBreakpointsOfPausedScript(scriptId);
+
+        // Important: For the logic that verifies if a user breakpoint is set in the paused location, we need to resolve pending breakpoints, and commit them, before
+        // using committedBreakpointsByUrl for our logic.
+        let anyPendingBreakpointsResolved = await this.resolvePendingBreakpointsOfPausedScript(pausedLocation.scriptId);
+
+        const pausedScriptUrl = this.getPausedScriptUrlFromId(pausedLocation.scriptId);
+        // Important: We need to get the commited breakpoints only after all the pending breakpoints for this file have been resolved. If not this logic won't work
+        const commitedBps = this._chromeDebugAdapter.committedBreakpointsByUrl.get(pausedScriptUrl);
+        const anyBreakpointsAtPausedLocation = commitedBps.filter(bp =>
```
rakatyal: We populate the committedBreakpointsByUrl on breakpointResolved right? Can there be a situation where we hit the stopOnEntry breakpoint and a user breakpoint on (1,1) hasn't been resolved yet?

src/chrome/breakOnLoadHelper.ts
```diff
@@ -122,16 +125,24 @@ export class BreakOnLoadHelper {
...
         let shouldContinue = true;
-        let anyPendingBreakpointsResolved = await this.resolvePendingBreakpointsOfPausedScript(scriptId);
+
+        // Important: For the logic that verifies if a user breakpoint is set in the paused location, we need to resolve pending breakpoints, and commit them, before
+        // using committedBreakpointsByUrl for our logic.
+        let anyPendingBreakpointsResolved = await this.resolvePendingBreakpointsOfPausedScript(pausedLocation.scriptId);
+
+        const pausedScriptUrl = this.getPausedScriptUrlFromId(pausedLocation.scriptId);
+        // Important: We need to get the commited breakpoints only after all the pending breakpoints for this file have been resolved. If not this logic won't work
+        const commitedBps = this._chromeDebugAdapter.committedBreakpointsByUrl.get(pausedScriptUrl);
```
rakatyal: Nit: *committed

src/chrome/breakOnLoadHelper.ts
digeff: We block on resolvePendingBreakpointsOfPausedScript which should set all the pending breakpoints before returning. My understanding is that that will guarantee all the breakpoints are set before we continue. @roblourens: You know a lot more about the code than us, do you think that is a correct assumption?

rakatyal: Hmm I think that's correct.

roblourens:
Could you add a test for this?

src/chrome/breakOnLoadHelper.ts
```diff
@@ -67,12 +66,16 @@ export class BreakOnLoadHelper {
         }
     }
 
+    private getPausedScriptUrlFromId(scriptId: string): string {
```
roblourens: Why just a "paused" script?

src/chrome/chromeDebugAdapter.ts
```diff
@@ -945,8 +945,8 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         }
 
         const committedBps = this._committedBreakpointsByUrl.get(script.url) || [];
-        if (committedBps.indexOf(params.breakpointId) === -1) {
-            committedBps.push(params.breakpointId);
+        if (committedBps.findIndex(committedBp => committedBp.breakpointId === params.breakpointId) === -1) {
```
roblourens: I prefer `if (!committedBps.find(...` instead of checking index -1

src/chrome/breakOnLoadHelper.ts
digeff: You are right. Renaming to getScriptUrlFromId

src/chrome/chromeDebugAdapter.ts
digeff: changed

PR #286, opened 2018-02-14
add pathMappingOverrides configuration

aharpervc:
Rather than implement wildcards, which seems too complex, I added a new "pathMappingOverrides" block to do a simple find/replace on mapped file paths. It seems to work for my case, which is to enable Sprockets 3 support:  https://github.com/Microsoft/vscode-chrome-debug/issues/598

Basically with this launch.json:

```json
"pathMapping": {
  "/assets": "${workspaceFolder}/app/assets/javascripts"
},
"pathMappingOverrides": {
  ".self": ""
}
```

I can debug assets like: `http://localhost:3000/assets/session.self.js` mapped to `~\my-rails-app\app\assets\javascripts\session.js`

### todo

- design approval?
- tests?

### DIFF
--- a/src/chrome/chromeUtils.ts
+++ b/src/chrome/chromeUtils.ts
@@ -9,7 +9,7 @@ import Crdp from '../../crdp/crdp';
 import * as utils from '../utils';
 import {ITarget} from './chromeConnection';
 
-export function targetUrlToClientPathByPathMappings(scriptUrl: string, pathMapping: any): string {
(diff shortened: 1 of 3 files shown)


roblourens:
I disagree that wildcards are too complex. It's essentially the same logic as sourceMapPathOverrides, which we already have. I would rather extend pathMappings to be more similar to sourceMapPathOverrides than add a new option which works differently. I just regret not seeing the need for this earlier, and naming pathMappings and sourceMapPathOverrides more similary.

aharpervc:
That's fine. I'm not married to this PR or this design. However, I don't have the capacity to crack open the existing wildcard logic so I went with the small, additive approach. You can close this PR if it's a dead end.

PR #287, opened 2018-02-15
Now we remove the break on load breakpoints after they are not needed any more

digeff:
Now we remove the break-on-load breakpoints after they are not needed any more.

This removes some flickering on Chrome that we see when reloading the page, because we are hitting and resuming from these breakpoints.

### DIFF
--- a/src/chrome/breakOnLoadHelper.ts
+++ b/src/chrome/breakOnLoadHelper.ts
@@ -9,13 +9,19 @@ import {ISetBreakpointResult, BreakOnLoadStrategy} from '../debugAdapterInterfac
 import Crdp from '../../crdp/crdp';
 import {ChromeDebugAdapter} from './chromeDebugAdapter';
 import * as ChromeUtils from './chromeUtils';
+import * as assert from 'assert';
+
+export interface UrlRegexAndFileSet {
+    urlRegex: string;
+    fileSet: Set<string>;
+}
 
 export class BreakOnLoadHelper {
 
     private _instrumentationBreakpointSet: boolean = false;
 
     // Break on load: Store some mapping between the requested file names, the regex for the file, and the chrome breakpoint id to perform lookup operations efficiently
-    private _stopOnEntryBreakpointIdToRequestedFileName = new Map<string, [string, Set<string>]>();
+    private _stopOnEntryBreakpointIdToRequestedFileName = new Map<string, UrlRegexAndFileSet>();
     private _stopOnEntryRequestedFileNameToBreakpointId = new Map<string, string>();
     private _stopOnEntryRegexToBreakpointId = new Map<string, string>();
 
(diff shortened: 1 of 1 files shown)


src/chrome/breakOnLoadHelper.ts
```diff
@@ -157,14 +157,31 @@ export class BreakOnLoadHelper {
...
-        hitBreakpoints.forEach(bp => {
-            if (!this._stopOnEntryBreakpointIdToRequestedFileName.has(bp)) {
+        for (let bp of hitBreakpoints) {
+            let regexAndFileNames = this._stopOnEntryBreakpointIdToRequestedFileName.get(bp);
+            if (!regexAndFileNames) {
                 notification.hitBreakpoints = [bp];
                 allStopOnEntryBreakpoints = false;
+            } else {
+                const normalizedMappedUrl = this._chromeDebugAdapter.fixPathCasing(mappedUrl);
+                if (regexAndFileNames[1].has(normalizedMappedUrl)) {
```
roblourens: We should really change this to an object with named properties, not an array. Don't know why I didn't make that comment earlier

roblourens: And this is for keeping track of when multiple files are covered by a single break on load breakpoint, right?

src/chrome/breakOnLoadHelper.ts
```diff
@@ -271,6 +289,11 @@ export class BreakOnLoadHelper {
         return result;
     }
 
+    // Removes a breakpoint on (0,0) for the files matching the given regex
```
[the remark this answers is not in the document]
digeff: Done

src/chrome/breakOnLoadHelper.ts
digeff: Awesome feedback. It made me realized that I was missing a bunch of code :)

src/chrome/breakOnLoadHelper.ts
```diff
@@ -157,14 +163,33 @@ export class BreakOnLoadHelper {
...
+                if (regexAndFileNames.fileSet.has(normalizedMappedUrl)) {
+                    regexAndFileNames.fileSet.delete(normalizedMappedUrl);
+                    assert(this._stopOnEntryRequestedFileNameToBreakpointId.delete(normalizedMappedUrl), `Expected to delete break-on-load information associated with url: ${normalizedMappedUrl}`);
+
+                    if (regexAndFileNames.fileSet.size === 0) {
+                        logger.log(`Stop on entry breakpoint hit for last remaining file. Removing: ${bp} created for: ${normalizedMappedUrl}`);
+                        await this.removeBreakpointById(bp);
+                        assert(this._stopOnEntryRegexToBreakpointId.delete(regexAndFileNames.urlRegex), `Expected to delete break-on-load information associated with regexp: ${regexAndFileNames.urlRegex}`);
+                    } else {
+                        logger.log(`Stop on entry breakpoint hit but still has remaining files. Keeping: ${bp} that was hit for: ${normalizedMappedUrl} because it's still needed for: ${Array.from(regexAndFileNames.fileSet.entries()).join(", ")}`);
```
rakatyal: This means that we will still see the flickering if this particular file/url is reloaded right(since we are not removing the breakpoint)? I guess there is no way to remove a particular match of a setBreakpointByUrl request?

digeff: It's technically possible to do it if we play with the regexp, but I don't think it's worth the risk or effort to do it.

rakatyal: Hmm all right.

rakatyal:
LGTM.

PR #288, opened 2018-02-15
Sends telemetry when we see unhandled exceptions.

changsi-an:
Sends telemetry when we see unhandled exceptions or unhandled promise rejections.

Also improves the logic for constructing telemetry payload for an
exception.

### DIFF
--- a/src/chrome/chromeDebugSession.ts
+++ b/src/chrome/chromeDebugSession.ts
@@ -33,6 +33,21 @@ export interface IChromeDebugSessionOpts extends IChromeDebugAdapterOpts {
     extensionName: string;
 }
 
+export type ExceptionType = "uncaughtException" | "unhandledRejection" | "generic";
+
+export interface  IExceptionTelemetryProperties {
+    // There is an issue on some clients and reportEvent only currently accept strings properties,
+    // hence all the following properties must be strings.
+    successful?: "true" | "false";
+    exceptionType?: ExceptionType;
+    exceptionMessage?: string;
+    exceptionName?: string;
+    exceptionStack?: string;
+    timeTakenInMilliseconds?: string;
+}
+
+export const ErrorTelemetryEventName = 'error';
+
 // A failed request can return either an Error, an error from Chrome, or a DebugProtocol.Message which is returned as-is to the client
 type RequestHandleError = Error | DebugProtocol.Message | IChromeError;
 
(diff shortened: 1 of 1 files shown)


changsi-an:
@digeff

src/chrome/chromeDebugSession.ts
```diff
@@ -202,6 +220,18 @@ export class ChromeDebugSession extends LoggingDebugSession {
     private sendUnknownCommandResponse(response: DebugProtocol.Response, command: string): void {
         this.sendErrorResponse(response, 1014, `[${this._extensionName}] Unrecognized request: ${command}`, null, ErrorDestination.Telemetry);
     }
+
+    private fillErrorDetails(properties: ITelemetryPropertiesPayload, e: any) {
+        if (e.message) {
+            properties.exceptionMessage = e.message || e.toString();
```
roblourens: `e.toString` will never execute here

src/chrome/chromeDebugSession.ts
```diff
@@ -82,14 +97,28 @@ export class ChromeDebugSession extends LoggingDebugSession {
             return errMsg;
         };
 
+        const reportErrorTelemetry = (err, exceptionType: ExceptionType)  => {
+            let properties: ITelemetryPropertiesPayload = {};
+            properties.successful = "false";
+            properties.exceptionType = exceptionType;
+            properties.timeTakenInMilliseconds = "unknown";
```
roblourens: Should this just be empty instead of a string that can't be parsed as a number?

src/chrome/chromeDebugSession.ts
```diff
@@ -33,6 +33,21 @@ export interface IChromeDebugSessionOpts extends IChromeDebugAdapterOpts {
     extensionName: string;
 }
 
+export type ExceptionType = "uncaughtException" | "unhandledRejection" | "generic";
+
+export interface  ITelemetryPropertiesPayload {
```
roblourens: Can this interface name be more specific? This isn't for all telemetry messages, just timings and errors.

src/chrome/chromeDebugSession.ts
```diff
@@ -33,6 +33,21 @@ export interface IChromeDebugSessionOpts extends IChromeDebugAdapterOpts {
     extensionName: string;
 }
 
+export type ExceptionType = "uncaughtException" | "unhandledRejection" | "generic";
```
roblourens: Is the `generic` type needed?

src/chrome/chromeDebugSession.ts
```diff
@@ -202,6 +220,18 @@ export class ChromeDebugSession extends LoggingDebugSession {
     private sendUnknownCommandResponse(response: DebugProtocol.Response, command: string): void {
         this.sendErrorResponse(response, 1014, `[${this._extensionName}] Unrecognized request: ${command}`, null, ErrorDestination.Telemetry);
     }
+
+    private fillErrorDetails(properties: ITelemetryPropertiesPayload, e: any) {
```
roblourens: Should this set exceptionType?

src/chrome/chromeDebugSession.ts
changsi-an: Good catch. Fixed.

src/chrome/chromeDebugSession.ts
changsi-an: Maybe not, caller can customize and set that

src/chrome/chromeDebugSession.ts
changsi-an: It's changed to empty.

src/chrome/chromeDebugSession.ts
changsi-an: Renamed

src/chrome/chromeDebugSession.ts
changsi-an: Yes. I just introduced a consumer now.

changsi-an:
Thanks. tslint is fixed.

PR #289, opened 2018-02-15
Add null check

digeff:
Sometimes we reject(null)


roblourens:
When was it null?

digeff:
For Promise<void> I usually do resolve(null)

roblourens:
This is for an error handler, so did you do reject(null)? If so there should be a real error message there.

If there's a specific case where you know this ends up as null, please try to fix it up so we don't have to guess at what "unhandled error: null" might be.

Issue #106, opened 2016-09-22
Bind unverified breakpoints when the sourcemap can only be found after the script is loaded


roblourens:
The existing Node debugger only binds breakpoints in source if their sourcemaps can be found from outDir/outFiles ahead of time, but there are some scenarios where sourcemaps can't be resolved ahead of time, or don't exist on disk.

To handle this, the adapter needs to hold on to unbound bp info, and when a script is loaded and new sourcemaps are resolved, it needs to bind those bps and send BreakpointEvents informing the frontend that the bps are now bound.

There are a few relevant scenarios in Node like with SystemJS and Babel dynamic require, but it's more important in browser scenarios where we can't resolve URL -> script until the script is loaded (without adding yet another config parameter). Used to work in Chrome until I started using configurationRequest, so I hope to create the full fix soon.

roblourens:
Leftover from #38

PR #290, opened 2018-02-15
Updated code so now we resolve pending breakpoints onScriptParsed instead onScriptPaused when using break on load

digeff:
The general idea of this code change is that now all the pending breakpoints will get resolved onScriptParsed. Because we don't know in which order things will be executed, we'll have a defer used as a semaphore. The break-on-load logic inside onScriptPaused won't continue until the associated onScriptParsed has finished resolving all the pending breakpoints.

We need to make this change because of files that start with a function declaration, 0,0 is inside the function, so the break-on-load breakpoint is sometimes never hit, so then even after we refresh the page, no breakpoints are hit in that file because they are all still pending.

Is there a better way to solve this issue?

Some specific question:
I had to remove: "&& !pendingBP.bpsSet?" from line 697 to make this work. I'm not sure how that code was being used. It removing that code the right thing to do?
How should I handle the refresh case? Do we need to do anything special for that case? Should I call      this._scriptIdToBreakpointsAreResolvedDefer.clear(); inside clearTargetContext?

### DIFF
--- a/src/chrome/breakOnLoadHelper.ts
+++ b/src/chrome/breakOnLoadHelper.ts
@@ -74,7 +74,7 @@ export class BreakOnLoadHelper {
 
             const pausedScriptId = notification.callFrames[0].location.scriptId;
             // Now we wait for all the pending breakpoints to be resolved and then continue
-            await this._chromeDebugAdapter.breakpointsResolvedDefer(pausedScriptId).promise;
+            await this._chromeDebugAdapter.getBreakpointsResolvedDefer(pausedScriptId).promise;
             return true;
         }
         return false;
@@ -90,7 +90,7 @@ export class BreakOnLoadHelper {
 
         // Important: For the logic that verifies if a user breakpoint is set in the paused location, we need to resolve pending breakpoints, and commit them, before
         // using committedBreakpointsByUrl for our logic.
-        await this._chromeDebugAdapter.breakpointsResolvedDefer(pausedLocation.scriptId).promise;
+        await this._chromeDebugAdapter.getBreakpointsResolvedDefer(pausedLocation.scriptId).promise;
 
         const pausedScriptUrl = this.getScriptUrlFromId(pausedLocation.scriptId);
         // Important: We need to get the committed breakpoints only after all the pending breakpoints for this file have been resolved. If not this logic won't work
@@ -98,7 +98,7 @@ export class BreakOnLoadHelper {
         const anyBreakpointsAtPausedLocation = committedBps.filter(bp =>
             bp.actualLocation.lineNumber === pausedLocation.lineNumber && bp.actualLocation.columnNumber === pausedLocation.columnNumber).length > 0;
 
-        // If there were any breakpoints at (1,1) we shouldn't continue
+        // If there were any breakpoints at this location (Which generally should be (1,1)) we shouldn't continue
         if (anyBreakpointsAtPausedLocation) {
             // Here we need to store this information per file, but since we can safely assume that scriptParsed would immediately be followed by onPaused event
             // for the breakonload files, this implementation should be fine
--- a/src/chrome/breakOnLoadHelper.ts
+++ b/src/chrome/breakOnLoadHelper.ts
@@ -52,55 +52,8 @@ export class BreakOnLoadHelper {
         return this._instrumentationBreakpointSet;
     }
 
-    /**
-     * Checks and resolves the pending breakpoints of a script given it's source. If any breakpoints were resolved returns true, else false.
-     * Used when break on load active, either through Chrome's Instrumentation Breakpoint API or the regex approach
-     */
-    private async resolvePendingBreakpoints(source: string): Promise<boolean> {
-        const normalizedSource = this._chromeDebugAdapter.fixPathCasing(source);
-        const pendingBreakpoints = this._chromeDebugAdapter.pendingBreakpointsByUrl.get(normalizedSource);
-        // If the file has unbound breakpoints, resolve them and return true
-        if (pendingBreakpoints !== undefined) {
-            await this._chromeDebugAdapter.resolvePendingBreakpoint(pendingBreakpoints);
-            if (!this._chromeDebugAdapter.pendingBreakpointsByUrl.delete(normalizedSource)) {
-                logger.log(`Expected to delete ${normalizedSource} from the list of pending breakpoints, but it wasn't there`);
-            }
-            return true;
-        } else {
-            // If no pending breakpoints, return false
-            return false;
-        }
-    }
-
     private getScriptUrlFromId(scriptId: string): string {
         return this._chromeDebugAdapter.scriptsById.get(scriptId).url;
-    }
-
-    /**
-     * Checks and resolves the pending breakpoints given a script Id. If any breakpoints were resolved returns true, else false.
-     * Used when break on load active, either through Chrome's Instrumentation Breakpoint API or the regex approach
-     */
-    private async resolvePendingBreakpointsOfPausedScript(scriptId: string): Promise<boolean> {
-        const pausedScriptUrl = this.getScriptUrlFromId(scriptId);
-        const sourceMapUrl = this._chromeDebugAdapter.scriptsById.get(scriptId).sourceMapURL;
-        const mappedUrl = await this._chromeDebugAdapter.pathTransformer.scriptParsed(pausedScriptUrl);
-        let breakpointsResolved = false;
-
-        let sources = await this._chromeDebugAdapter.sourceMapTransformer.scriptParsed(mappedUrl, sourceMapUrl);
-
-        // If user breakpoint was put in a typescript file, pendingBreakpoints would store the typescript file in the mapping, so we need to hit those
-        if (sources) {
-            for (let source of sources) {
-                let anySourceBPResolved = await this.resolvePendingBreakpoints(source);
-                // If any of the source files had breakpoints resolved, we should return true
-                breakpointsResolved = breakpointsResolved || anySourceBPResolved;
-            }
-        }
-        // If sources is not present or user breakpoint was put in a compiled javascript file
-        let scriptBPResolved = await this.resolvePendingBreakpoints(mappedUrl);
-        breakpointsResolved = breakpointsResolved || scriptBPResolved;
-
-        return breakpointsResolved;
     }
 
     /**
@@ -120,8 +73,8 @@ export class BreakOnLoadHelper {
             // This is fired when Chrome stops on the first line of a script when using the setInstrumentationBreakpoint API
 
             const pausedScriptId = notification.callFrames[0].location.scriptId;
-            // Now we should resolve all the pending breakpoints and then continue
-            await this.resolvePendingBreakpointsOfPausedScript(pausedScriptId);
+            // Now we wait for all the pending breakpoints to be resolved and then continue
+            await this._chromeDebugAdapter.breakpointsResolvedDefer(pausedScriptId).promise;
             return true;
         }
         return false;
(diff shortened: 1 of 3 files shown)


roblourens:
Is there no guarantee that a scriptParsed event will come before a pause event on that script? Have you ever seen them come in the wrong order or is it just being careful?

> I had to remove: "&& !pendingBP.bpsSet?" from line 697 to make this work. I'm not sure how that code was being used. It removing that code the right thing to do?

That code is for the case where we set a BP and believe we have the correct sourcemapped location for it, but the script is not yet loaded. That exact check was added recently in https://github.com/Microsoft/vscode-chrome-debug-core/commit/c12522435a6c11a5b5c4043612694522ff9063d7 and the bug was quite bad so the check should stay. But I suspect that when I fixed that, I actually broke something else which is what `bpsSet` was originally added for - the case when we think we have the correct location, but actually don't (so the pending BP still needs to be resolved) but I'll look at that later. Why did you have to remove it? I don't think your pending BPs should have that flag set.

> How should I handle the refresh case? Do we need to do anything special for that case?

I'm not sure, what do you mean by "refresh case" exactly?

digeff:
After discussing this with Dhanvi, we decided not to take this change.

digeff:
@roblourens: After discussing this in more depth, we decided to merge this, although we are not sure if we want to merge it for this release, or the next one. Please help us get this into a state where it's ready to merge, and after we get to that state we'll decide whether to merge it in this release, or in the next release

digeff:
Is there no guarantee that a scriptParsed event will come before a pause event on that script? Have you ever seen them come in the wrong order or is it just being careful? --> I'm assuming that you are asking why do we need this "waiting" logic. If that's the case, onScriptParsed and onScriptPaused do always arrive in the correct order. The issue is that we are not blocking the thread until all the work gets done. The resolved breakpoints are set in a non-blocking async manner, so it's possible that onScriptPaused executes while parts of onScriptParsed are still executing (in particular before we resolve all the breakpoints).

digeff:
I'm not sure, what do you mean by "refresh case" exactly? --> I'm not sure whether we need some special logic on clearTargetContext or something like that. What happens if we navigate or we refresh while we are in the middle of resolving breakpoints onScriptParsed and we navigate to the same page, or a different page? Can we have two onScriptParsed handlers for the same url been executed at the same time (one for before the refresh, one for after)? How do we handle things properly in that case?

digeff:
@roblourens I've updated the PR to do something reasonable with bpSet, rather than just remove it :).

src/chrome/breakOnLoadHelper.ts
```diff
@@ -137,16 +90,16 @@ export class BreakOnLoadHelper {
...
         const pausedScriptUrl = this.getScriptUrlFromId(pausedLocation.scriptId);
         // Important: We need to get the committed breakpoints only after all the pending breakpoints for this file have been resolved. If not this logic won't work
-        const committedBps = this._chromeDebugAdapter.committedBreakpointsByUrl.get(pausedScriptUrl);
+        const committedBps = this._chromeDebugAdapter.committedBreakpointsByUrl.get(pausedScriptUrl) || [];
         const anyBreakpointsAtPausedLocation = committedBps.filter(bp =>
             bp.actualLocation.lineNumber === pausedLocation.lineNumber && bp.actualLocation.columnNumber === pausedLocation.columnNumber).length > 0;
 
-        // If there were any pending breakpoints resolved and any of them was at (1,1) we shouldn't continue
-        if (anyPendingBreakpointsResolved && anyBreakpointsAtPausedLocation) {
+        // If there were any breakpoints at (1,1) we shouldn't continue
```
roblourens: "at (1,1)" is this still accurate? It can be anywhere now?

src/chrome/chromeDebugAdapter.ts
```diff
@@ -646,75 +655,88 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
-            if (this._earlyScripts) {
-                this._earlyScripts.push(script);
-            } else {
-                await this.sendLoadedSourceEvent(script);
-            }
-        });
-
-        if (script.url) {
-            script.url = utils.fixDriveLetter(script.url);
+    public breakpointsResolvedDefer(scriptId: string): PromiseDefer<void> {
```
roblourens: What's the verb in this method name, should it be a "get"?

src/chrome/chromeDebugAdapter.ts
```diff
@@ -646,75 +655,88 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
+            return newValue;
         }
+    }
+
+    protected async onScriptParsed(script: Crdp.Debugger.ScriptParsedEvent): Promise<void> {
+        const breakpointsAreResolvedDefer = this.breakpointsResolvedDefer(script.scriptId);
+        try {
+            this.doAfterProcessingSourceEvents(async () => { // This will block future 'removed' source events, until this processing has been completed
+                if (typeof this._columnBreakpointsEnabled === 'undefined') {
+                    await this.detectColumnBreakpointSupport(script.scriptId).then(async () => {
```
roblourens: Don't mix async/await and promises when possible

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1316,7 +1341,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         };
     }
 
-    private targetBreakpointResponsesToClientBreakpoints(url: string, responses: ISetBreakpointResult[], requestBps: DebugProtocol.SourceBreakpoint[], ids?: number[]): DebugProtocol.Breakpoint[] {
+    private targetBreakpointResponsesToClientBreakpoints(url: string, responses: ISetBreakpointResult[], requestBps: DebugProtocol.SourceBreakpoint[], ids?: number[]): BreakpointSetResult[] {
```
roblourens: method name isn't accurate with this change

src/chrome/chromeDebugAdapter.ts
```diff
@@ -646,75 +655,88 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
+                    breakpointsAreResolvedDefer.resolve();
+                }
+            };
+
+            const sourceMapsP = this._sourceMapTransformer.scriptParsed(mappedUrl, script.sourceMapURL).then(async sources => {
+                if (this._hasTerminated) {
+                    return undefined;
+                }
+
+                if (sources) {
```
roblourens: When is this undefined?

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1140,10 +1162,13 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
                     });
 
                     // Return the setBP request, no matter how long it takes. It may take awhile in Node 7.5 - 7.7, see https://github.com/nodejs/node/issues/11589
-                    return setBreakpointsPFailOnError.then(body => {
+                    return setBreakpointsPFailOnError.then(setBpResultBody => {
+                        const body = { breakpoints: setBpResultBody.breakpoints.map(setBpResult => setBpResult.breakpoint) };
                         if (body.breakpoints.every(bp => !bp.verified)) {
+                            // If all breakpoints are set, we mark them as set. If not, we mark them as un-set so they'll be set
+                            const areAllSet = setBpResultBody.breakpoints.every(setBpResult => setBpResult.isSet);
```
roblourens: I think this changes the way it works - bpsSet was true if the bp was attempted to be set - not that it was bound. The check below `response.breakpointId !== undefined` is only true when the bp is bound.

roblourens:
I'm not sure refresh is an issue, because of `doAfterProcessingSourceEvents` and once pending breakpoints are resolved, they don't need to become pending again because once the script has been loaded once during the lifetime of the adapter, we have set the bp in the correct location.

This is a complicated change and I'm not sure I follow the whole thing. Take a look at the feedback and let's try to have a call about it in the next day or two.

src/chrome/breakOnLoadHelper.ts
digeff: // If there were any breakpoints at this location (Which generally should be (1,1)) we shouldn't continue

src/chrome/chromeDebugAdapter.ts
digeff: getBreakpointsResolvedDefer

src/chrome/chromeDebugAdapter.ts
digeff: if (typeof this._columnBreakpointsEnabled === 'undefined') {
                    await this.detectColumnBreakpointSupport(script.scriptId);
                    await this.sendInitializedEvent();
                }

src/chrome/chromeDebugAdapter.ts
digeff: targetBreakpointResponsesToBreakpointSetResults

src/chrome/chromeDebugAdapter.ts
digeff: This code was added for this: https://github.com/Microsoft/vscode-chrome-debug-core/issues/106 in https://github.com/Microsoft/vscode-chrome-debug-core/commit/1bf19e826b389f4033faa76287921600cf03ca16 so I'm assuming that the scenario is:
"sourcemaps can't be resolved ahead of time, or don't exist on disk."

src/chrome/chromeDebugAdapter.ts
digeff: My understanding was that Debugger.setBreakpointByUrl always returns the breakpointId if the parameters are valid even if the breakpoint is unbound, and it also returns the actualLocation if the breakpoint was bound.

digeff:
I'm adding some more awaits because if not the second call to breakpointsAreResolvedDefer.resolve() might be called to early...

roblourens:
Please add some integration tests for this too

src/chrome/chromeDebugAdapter.ts
roblourens: Sorry, you're correct.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1350,29 +1373,31 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
                 }
 
                 if (!response.actualLocation) {
-                    return <DebugProtocol.Breakpoint>{
+                    // If we don't have an actualLocation nor a breakpointId this is a pseudo-breakpoint because we are using break-on-load
+                    // so we mark the breakpoint as not set, so i'll be set after we load the actual script that has the breakpoint
+                    return { isSet: response.breakpointId !== undefined, breakpoint: <DebugProtocol.Breakpoint>{
```
roblourens: Nitpick, please format these like 

```
{
	isSet...
	breakpoint: {
		...
	}
}
```

as I didn't notice there was a second object there at first.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -646,75 +655,86 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
-        };
+            const resolvePendingBPs = async (source: string) => {
+                source = source && this.fixPathCasing(source);
+                const pendingBP = this._pendingBreakpointsByUrl.get(source);
+                if (pendingBP && !pendingBP.bpsSet) {
+                    await this.resolvePendingBreakpoint(pendingBP);
+                    this._pendingBreakpointsByUrl.delete(source);
+                    breakpointsAreResolvedDefer.resolve();
+                } else {
+                    breakpointsAreResolvedDefer.resolve();
```
roblourens: You can just pull these two out of the if/else

src/chrome/chromeDebugAdapter.ts
digeff: Changed

src/chrome/chromeDebugAdapter.ts
digeff: I didn't noticed that, thanks :)... It's actually a bug to do these here, so I'm removing both of them....

src/chrome/chromeDebugAdapter.ts
```diff
@@ -646,75 +655,84 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
-                }
-            }
+                await this.resolveSkipFiles(script, mappedUrl, sources);
+            });
 
-            return this.resolveSkipFiles(script, mappedUrl, sources);
-        });
+            if (this._initialSourceMapsP) {
+                this._initialSourceMapsP = <Promise<any>>Promise.all([this._initialSourceMapsP, sourceMapsP]);
+                await sourceMapsP;
```
roblourens: Should this always be awaited outside the if block? And why does it need to be?

roblourens:
Left one more comment but you should have permission to merge it, please do that when you think it's ready. Let me know if it doesn't work, I'll do it.

src/chrome/chromeDebugAdapter.ts
digeff: You are right, it should be outside.

We want this so we don't call breakpointsAreResolvedDefer.resolve(); until this has finished for sure. If we call breakpointsAreResolvedDefer.resolve(); too early that will cause issues.

digeff:
@roblourens Any more feedback?

PR #294, opened 2018-02-27
Now we report the times of the different tasks that happen until we launch and show the user page

digeff:
We add a lot of telemetry to track where the time is going when we launch chrome

Related to this PR: https://github.com/Microsoft/vscode-chrome-debug/pull/604


digeff:
@roblourens @changsi-an

src/chrome/chromeDebugSession.ts
```diff
@@ -202,6 +207,17 @@ export class ChromeDebugSession extends LoggingDebugSession {
     private sendUnknownCommandResponse(response: DebugProtocol.Response, command: string): void {
         this.sendErrorResponse(response, 1014, `[${this._extensionName}] Unrecognized request: ${command}`, null, ErrorDestination.Telemetry);
     }
+
+    public reportTimingsUntilUserPage(): void {
```
changsi-an: Who calls this?

digeff:
The EvenEmitter is a good idea, I'm refactoring the code to use that.

src/chrome/chromeConnection.ts
```diff
@@ -29,7 +30,7 @@ export interface ITarget {
 }
 
 export type ITargetFilter = (target: ITarget) => boolean;
-export interface ITargetDiscoveryStrategy {
+export interface ITargetDiscoveryStrategy extends ObservableEvents {
```
changsi-an: Should it extend ObservableEvents ? Or let the concrete class implement both?

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,128 @@
...
+
+interface MilestoneReachedEventArguments {
+    milestoneName: string;
+}
+
+export interface ObservableEvents {
+    Events: EventEmitter;
+}
+
+export class StepStartedEventsEmitter extends EventEmitter {
```
changsi-an: I feel that this class already has a verb `start` in it. So repeat that in the class name might be a little off?  Since it's not all about `start` (correct me if I am wrong).

Would it be more like a `StepProgressEventsEmitter` ?

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,128 @@
...
+    constructor(public readonly NestedEmitters: [EventEmitter] = [] as [EventEmitter]) {
+        super();
+    }
+
+    public emitStepStarted(stepName: string): void {
+        this.emit(stepStartedEventName, { stepName: stepName } as StepStartedEventArguments);
+    }
+
+    public emitMilestoneReached(milestoneName: string): void {
```
changsi-an: Should we make a correspondence between two method? One is `start`, the other is `finish/compete`? instead of introducing a new concept of `milestone` (which in some sense might collide with the concept of `step` )

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,128 @@
...
+interface StepStartedEventArguments {
+    stepName: string;
+}
+
+interface MilestoneReachedEventArguments {
+    milestoneName: string;
+}
+
+export interface ObservableEvents {
+    Events: EventEmitter;
```
changsi-an: Should this type be `StepStartedEventsEmitter` ?

src/executionTimingsReporter.ts
changsi-an: And who calls this now?

src/executionTimingsReporter.ts
digeff: Done

src/chrome/chromeConnection.ts
digeff: Done

src/executionTimingsReporter.ts
digeff: I think it's better to promise only the minimum strictly necessary in the interface/contract, so then the implementer is free to solve this in whichever way seems best. If this property is an EventEmitter the code will work. We don't currently need it to be any particular implementation.

src/executionTimingsReporter.ts
digeff: This gets called from vscode-chrome-debug in the other PR. Let's sync offline about your concern...

src/executionTimingsReporter.ts
```diff
@@ -31,13 +41,13 @@ export class StepProgressEventsEmitter extends EventEmitter {
         this.emit(milestoneReachedEventName, { milestoneName: milestoneName } as MilestoneReachedEventArguments);
     }
 
-    private subscribeToAllNestedEmitters(event: string | symbol, listener: Function): void {
+    private subscribeToAllNestedEmitters(event: string, listener: Function): void {
         for (const nestedEventEmitter of this._nestedEmitters) {
-            nestedEventEmitter.on(event, listener);
+            nestedEventEmitter.on(event as any, listener as any);
```
changsi-an: Wouldn't an 'event` parameter be just a string?

digeff: the parameter is something like 'stepStarted'|'milestoneReached'|'navigatedToUserRequestedUrl'

PR #295, opened 2018-03-01
Enable telemetry measurement for scriptParsed event.

changsi-an:


src/chrome/chromeDebugAdapter.ts
```diff
@@ -584,7 +588,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
         this._expectingStopReason = undefined;
 
-        smartStepP.then(should => {
+        await smartStepP.then(should => {
```
changsi-an: This change is intentional. We should take regard of the time spent on this operation when we measure `Debugger.paused` event.

digeff:
👍

changsi-an:
@msft-cwells  Yes tested and saw the telemetry as expected

PR #296, opened 2018-03-02
Fix some unexplained breakpoints behavior

digeff:
If the breakpointId is a number, these two types of key can get incorrectly mixed with each other.


digeff:
@roblourens

digeff:
@changsi-an

digeff:
@roblourens We want you to take a look before merging this, so we are retargeting this PR to master.

roblourens:
What's the problem exactly? I don't understand the description.

roblourens:
I think this is the wrong fix, the point of keeping unbound bps in this set of handles is that they are the same thing as the bound BPs, and the .set call is recording the change from an unbound BP to a bound BP. I think a better thing to do is to change the unbound ID to not conflict with a real id, like by adding `_unbound` to the number or something.

digeff:
Applied feedback

PR #300, opened 2018-03-06
Batch sending telemetry for some of the event notifications.

changsi-an:
Batch-sending telemetry is necessary because some notifications from debuggee might be too frequent for the telemetry infrastructure to handle. Take VS as an example, it throttles telemetry events if there are two many within a short period of time. We will lose some important telemetry when that happens.
The BatchTelemetryReporter will cache the received telemetry events and sends them out every 10 seconds.

This PR also refactors some code and migrates some general telemetry handling code from DebugSession.ts to utils.ts or telemetry.ts.

This PR also enables telemetry for debugger.paused and debugger.onScriptParsed notification events.

This PR retargets the change to master branch.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -21,7 +21,7 @@ import {StoppedEvent2, ReasonType} from './stoppedEvent';
 
 import * as errors from '../errors';
 import * as utils from '../utils';
-import {telemetry} from '../telemetry';
+import {telemetry, BatchTelemetryReporter, IExecutionResultTelemetryProperties} from '../telemetry';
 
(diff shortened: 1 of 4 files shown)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -567,7 +600,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
         this._expectingStopReason = undefined;
 
-        smartStepP.then(should => {
+        await smartStepP.then(should => {
```
changsi-an: I add it to include the time spent in this operation in the overall measurement of onPaused event. Apart from the measurement function, no one seems to actively wait on the promise returned from this method.

PR #301, opened 2018-03-07
Now we report the times of the different tasks that happen until we launch and show the user page

digeff:
Now we report the times of the different tasks that happen until we launch and show the user page


test/chrome/chromeDebugAdapter.test.ts
```diff
@@ -57,11 +58,14 @@ suite('ChromeDebugAdapter', () => {
         mockChromeConnection
             .setup(x => x.attach(It.isValue(undefined), It.isValue(ATTACH_FAIL_PORT), It.isValue(undefined), It.isValue(undefined)))
             .returns(() => utils.errP('Testing attach failed'));
-        mockChromeConnection
+            mockChromeConnection
```
roblourens: Fix indentation

test/chrome/chromeDebugAdapter.test.ts
```diff
@@ -57,11 +58,14 @@ suite('ChromeDebugAdapter', () => {
...
             .setup(x => x.isAttached)
             .returns(() => false);
         mockChromeConnection
             .setup(x => x.onClose(It.isAny()));
+        mockChromeConnection
+            .setup(x => x.Events)
+            .returns(() => new StepProgressEventsEmitter);
```
roblourens: Missing ()?

src/chrome/chromeDebugAdapter.ts
```diff
@@ -152,11 +153,15 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
     private _batchTelemetryReporter: BatchTelemetryReporter;
 
-    public constructor({ chromeConnection, lineColTransformer, sourceMapTransformer, pathTransformer, targetFilter, enableSourceMapCaching }: IChromeDebugAdapterOpts, session: ChromeDebugSession) {
+    public readonly Events: StepProgressEventsEmitter;
```
roblourens: Why capital E?

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,139 @@
+import { HighResTimer, calculateElapsedTime } from "./utils";
```
roblourens: Copyright header on new file

roblourens: And single quotes

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export interface StepStartedEventArguments {
+    stepName: string;
+}
+
+export interface MilestoneReachedEventArguments {
+    milestoneName: string;
+}
+
+export interface ObservableEvents<T> { // Normally T will be an EventEmitter
```
roblourens: Then you can say `T extends EventEmitter`

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    public on(event: string, listener: Function): this {
+        super.on(event, listener);
+        this.subscribeToAllNestedEmitters(event, listener);
+        return this;
+    }
+}
+
+/* Use to track the time executing each step during launch
```
roblourens: Is this usage comment necessary? It's incomplete since it doesn't say how to create Events or a reporter, and things like this quickly get out of date. Since this already includes code that uses these objects, I'm not sure this is needed.

src/chrome/chromeDebugSession.ts
```diff
@@ -131,6 +140,8 @@ export class ChromeDebugSession extends LoggingDebugSession {
                     reportFailure(e);
                 }
                 this.failedRequest(request.command, response, e);
+            } finally {
+                this.Events.emitStepStarted(`WaitingAfter.ClientRequest.${request.command}`);
```
roblourens: Would it make more sense to emit that a step is complete, and have the emitter create the WaitingAfter events?

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,139 @@
...
+        this.Events.emitStepStarted("WaitingAfter.ClientRequest.setBreakpoints");
+        reporter.generateReport() // Returns the report. Do not call any more methods after this
+
+    The report/telemetry generated looks like this:
+        All	5271.022216 // Time since the reporter is created (which is basically when we launched the debug adapter itself) until we navigated to the user's web page
+        BeforeFirstStep	7.833247 // Time since the reporter is created (which is basically when we launched the debug adapter itself) until we get the first request
+        ClientRequest.initialize	[1.261372] // Time we spent processing the initialize request
+        WaitingAfter.ClientRequest.initialize	[74.939561] // Time we spent after processing the initialize request for the client to send us another request
+        ClientRequest.launch	[2.748096] // Time we spent processing the first part of the launch request until we actually start launching the target debugee .exe
+        LaunchTarget.LaunchExe	10.276666 // The time it takes to spawn the  the target debugee .exe (We launch the .exe but we don't wait for it)
```
roblourens: Why are some arrays and some not?

src/chrome/chromeDebugSession.ts
```diff
@@ -47,9 +48,14 @@ function isChromeError(e: RequestHandleError): e is IChromeError {
...
-export class ChromeDebugSession extends LoggingDebugSession {
-    private _debugAdapter: IDebugAdapter;
+export class ChromeDebugSession extends LoggingDebugSession implements ObservableEvents<StepStartedEventsEmitter & NavigatedToUserRequestedUrlEventsEmitter> {
+    private _debugAdapter: IDebugAdapter & ObservableEvents<StepStartedEventsEmitter & NavigatedToUserRequestedUrlEventsEmitter>;
     private _extensionName: string;
+    public readonly Events: StepProgressEventsEmitter;
+    private reporter = new ExecutionTimingsReporter();
+    private haveLaunchExecutionTimingsBeenReported = false;
+
+    public static readonly NavigatedToUserRequestedUrlEventName = 'navigatedToUserRequestedUrl';
```
roblourens: Where is this fired from? And it's not clear that we need interfaces for every event that will be fired.

src/executionTimingsReporter.ts
```diff
@@ -0,0 +1,139 @@
...
+export class StepProgressEventsEmitter extends EventEmitter {
+    constructor(private readonly _nestedEmitters: [StepStartedEventsEmitter] = [] as [StepStartedEventsEmitter]) {
+        super();
+    }
+
+    public emitStepStarted(stepName: string): void {
+        this.emit(stepStartedEventName, { stepName: stepName } as StepStartedEventArguments);
+    }
+
+    public emitMilestoneReached(milestoneName: string): void {
```
roblourens: Is this used anywhere?

test/chrome/chromeDebugAdapter.test.ts
digeff: Changed

test/chrome/chromeDebugAdapter.test.ts
digeff: Changed

src/chrome/chromeDebugAdapter.ts
digeff: I think I was thinking in C#... I'll change it to lowercase...

src/executionTimingsReporter.ts
digeff: changed

src/executionTimingsReporter.ts
digeff: Changed

src/executionTimingsReporter.ts
digeff: I removed the usage, and I left the sample telemetry

src/executionTimingsReporter.ts
digeff: I forgot to update that comment.
"All" and milestones are single values. All the steps are now arrays.

src/chrome/chromeDebugSession.ts
digeff: This will get fired from onFrameNavigated in vscode-chrome-debug.

src/executionTimingsReporter.ts
digeff: This will be used in configurationDone in vscode-chrome-debug

src/chrome/chromeDebugSession.ts
digeff: I think that conceptually it does make more sense to emit when a step is completed.

In practice, some steps end sometimes in different locations due to multiple return statements or other strategies, and most steps usually begin in a single place, so it seemed simpler to track step beginnings rather than steps ends...

Ideally we should track both being and end with something like recordStep(() = doStepHere());. But that requires a much bigger code change, so I think that the simpler approach is better at this time. What do you think?

roblourens: Or, the emitter can fire some event that says the step is complete, and the reporter could write that out as a WaitingAfter step if it wants to. Would that be more natural? If not you can ignore this feedback.

src/executionTimingsReporter.ts
digeff: Sorry, that was an older version of the code. Now T is just an interface, so we cannot make it extend EventEmitter. I changed this for:
// T is an interface that declares the on methods (listeners) that we can subscribe to

src/chrome/chromeDebugSession.ts
roblourens: I don't think it should know about "ClientRequests". I see that some steps just end when the next one starts, and these have an explicit end.

Is it a problem that steps end with the next one? What about overlapping requests from the client?

Also, why do we need to track the timings here, doesn't the reportTelemetry method with the callback do that?

src/executionTimingsReporter.ts
roblourens: That will get out of date too. Is it helpful?

digeff: I think it's helpful to have a reference to what kind of telemetry we expect. If you don't like having this in the code, I can move this to a OneNote or something

src/executionTimingsReporter.ts
digeff: Some steps like Attach.RequestDebuggerTargetsInformation or setBreakpoints can happen multiple times. If that happens we send [<timeTakenInFirstOcurrence>, <timeTakenInSecondOcurrence>]. For simplicity we assume that all "steps" can have multiple ocurrences

src/chrome/chromeDebugSession.ts
digeff: I'm sorry, I should've sent the two PRs at the same time to give you the full context.
This PR works in conjunction with: https://github.com/Microsoft/vscode-chrome-debug/pull/610
That is where we turn it off, etc...

src/chrome/chromeDebugSession.ts
roblourens: I don't think it should be referenced from here then. Navigating is a chrome-debug-specific concept, not something that -core needs to know about.

src/chrome/chromeDebugSession.ts
roblourens: I still don't see how it's turned off in chrome-debug. Also I don't think it makes sense to start logging from -core and only turn it off from the subclass. Then it will log forever for anything besides chrome-debug that doesn't reach up and turn it off.

If this is all for chrome-debug, does any of it make sense for node-debug or any other consumers? If not, we need to rethink this approach.

I also still don't understand why the telemetry that times each request isn't sufficient for the part of this that measures requests. It's the right model for this sort of thing. setBreakpoints, and setExceptionBreakpoints won't come necessarily come one at a time as steps. They could overlap, and then I think your timings would be confusing.

src/executionTimingsReporter.ts
digeff: Removed

src/chrome/chromeDebugSession.ts
```diff
@@ -138,12 +148,23 @@ export class ChromeDebugSession extends LoggingDebugSession {
     // { command: request.command, type: request.type };
     private async reportTelemetry(eventName: string, propertiesSpecificToAction: {[property: string]: string}, action: (reportFailure: (failure: any) => void) => Promise<void>): Promise<void> {
         const startProcessingTime = process.hrtime();
+        const startTime = Date.now();
+        const isSequentialRequest = eventName === "clientRequest/initialize" || eventName === "clientRequest/launch";
```
roblourens: attach?

src/chrome/chromeDebugAdapter.ts
```diff
@@ -314,7 +319,8 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             }
         */
         telemetry.reportEvent('debugStarted', { request: 'attach', args: Object.keys(args) });
-        return this.doAttach(args.port, args.url, args.address, args.timeout, args.websocketUrl, args.extraCRDPChannelPort);
+        await this.doAttach(args.port, args.url, args.address, args.timeout, args.websocketUrl, args.extraCRDPChannelPort);
+        this.events.emitMilestoneReached(ChromeDebugSession.FinishedStartingUpEventName);
```
roblourens: I guess for launch this will be called from chrome-debug. For consistency, let's remove this and emit this from chrome-debug too.

src/chrome/chromeDebugSession.ts
digeff: changed

PR #303, opened 2018-03-13
Change the way error message is surfaced up to the UI.

changsi-an:
Please see the inline comment.

### DIFF
--- a/src/chrome/chromeDebugSession.ts
+++ b/src/chrome/chromeDebugSession.ts
@@ -204,11 +204,14 @@ export class ChromeDebugSession extends LoggingDebugSession implements IObservab
             return;
         }
 
-        const errMsg = isChromeError(error) ?
+        const errUserMsg = isChromeError(error) ?
             error.message + ': ' + error.data :
-            (error.stack || error.message);
+            (error.message || error.stack);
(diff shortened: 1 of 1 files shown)


src/chrome/chromeDebugSession.ts
```diff
@@ -204,19 +204,22 @@ export class ChromeDebugSession extends LoggingDebugSession implements IObservab
             return;
         }
 
-        const errMsg = isChromeError(error) ?
+        const errUserMsg = isChromeError(error) ?
             error.message + ': ' + error.data :
-            (error.stack || error.message);
+            (error.message || error.stack);
```
changsi-an: `error.message` is more user friendly than `error.stack`(which is a stack for the DA itself, not the user code). So when reporting an error to the UI, we let `error.message` take priority.

The original message is put to `errDiagnosticMsg`, and emitted by the Output event. So we are not losing anything with this code change. If user is interested, s/he can still find the callstack in the output pane.

karama78:
Je ne comprend pas l'anglais

thilognedierry

2018-03-13 1:09 GMT+00:00 Chance An <notifications@github.com>:

PR #307, opened 2018-03-16
Add versions to telemetry

digeff:
Add versions to telemetry

We'll use this to add the debug adapter versions, and the browser user agent to telemetry

The code to add the browser user agent is on chrome-debug added on this PR: https://github.com/Microsoft/vscode-chrome-debug/pull/621

### DIFF
--- a/src/chrome/chromeDebugSession.ts
+++ b/src/chrome/chromeDebugSession.ts
@@ -260,5 +260,5 @@ function logVersionInfo(): void {
     logger.log(`Adapter node: ${process.version} ${process.arch}`);
     const coreVersion = require('../../../package.json').version;
     logger.log('vscode-chrome-debug-core: ' + coreVersion);
-    telemetry.addCustomGlobalProperty( { "Versions.DebugAdapterCore": coreVersion });
+    telemetry.addCustomGlobalProperty( { 'Versions.DebugAdapterCore': coreVersion });
 }
--- a/src/chrome/chromeDebugSession.ts
+++ b/src/chrome/chromeDebugSession.ts
@@ -258,5 +258,7 @@ export class ChromeDebugSession extends LoggingDebugSession implements IObservab
 function logVersionInfo(): void {
     logger.log(`OS: ${os.platform()} ${os.arch()}`);
     logger.log(`Adapter node: ${process.version} ${process.arch}`);
-    logger.log('vscode-chrome-debug-core: ' + require('../../../package.json').version);
+    const coreVersion = require('../../../package.json').version;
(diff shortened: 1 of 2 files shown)


src/telemetry.ts
```diff
@@ -25,17 +26,59 @@ export interface ITelemetryReporter {
...
+        this._actionsQueue = this._actionsQueue.then(() => // We block the report event until all the addCustomGlobalProperty have finished
+            this._telemetryReporter.reportEvent(name, data));
+    }
+
+    public setupEventHandler(_sendEvent: (event: DebugProtocol.Event) => void): void {
+        this._telemetryReporter.setupEventHandler(_sendEvent);
+    }
+
+    public addCustomGlobalProperty(additionalGlobalPropertiesPromise: Promise<any> | any): void {
+        this._actionsQueue = this._actionsQueue.then(async () => {
```
roblourens: Here's a suggestion for a different implementation. It might get the intent across a little better since yours is acting like a queue here but doesn't really need to be. But it's not a big difference so I leave it up to you which one to use.

```ts
const reportedPropertyP = additionalGlobalPropertiesPromise.then(prop => this._telemetryReporter.addCustomGlobalProperty, err => this.reportErrorWhileWaitingForProperty(err);

this._actionsQueue = Promise.join(
	this._actionsQueue,
	reportedPropertyP)
```

src/chrome/chromeDebugSession.ts
```diff
@@ -258,5 +258,7 @@ export class ChromeDebugSession extends LoggingDebugSession implements IObservab
 function logVersionInfo(): void {
     logger.log(`OS: ${os.platform()} ${os.arch()}`);
     logger.log(`Adapter node: ${process.version} ${process.arch}`);
-    logger.log('vscode-chrome-debug-core: ' + require('../../../package.json').version);
+    const coreVersion = require('../../../package.json').version;
+    logger.log('vscode-chrome-debug-core: ' + coreVersion);
+    telemetry.addCustomGlobalProperty( { "Versions.DebugAdapterCore": coreVersion });
```
roblourens: Please use single quotes (trying to enforce this now, I need to set up the tslint rule)

src/telemetry.ts
digeff: I'm using this suggestion with Promise.all

src/chrome/chromeDebugSession.ts
digeff: Changed

src/telemetry.ts
roblourens: Ah that's what I meant.

PR #308, opened 2018-03-19
Now we canonicalize file paths to lowercase when on win32

digeff:
Now we canonicalize file paths to lowercase when on win32

We found some scenarios were breakpoints weren't bounding nor hitting because the client sent the webRoot with a different case than the paths when setBreakpoints was called. That resulted in
 UrlPathTransformer.getTargetPathFromClientPath not finding the proper target urls because we were comparing file paths with different cases, so _clientPathToTargetUrl.get wasn't finding the expected result.

Canonicalizing to lowercase on windows for file paths fixes this issue.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -118,8 +118,19 @@ export function canonicalizeUrl(urlOrPath: string): string {
 
     urlOrPath = stripTrailingSlash(urlOrPath);
     urlOrPath = fixDriveLetterAndSlashes(urlOrPath);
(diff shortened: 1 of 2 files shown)


roblourens:
This sounds like the same problem as what I just fixed in https://github.com/Microsoft/vscode-chrome-debug-core/commit/d9639e7918ed9b1cce53017bbfcce68d5da53f85

Did you see that with a version of -core that includes that commit?

digeff:
I didn't. The new version fixes the scenario I had. Abandoning this PR.

PR #314, opened 2018-03-22
Introduces a telemetry collector.

changsi-an:
so a request handling logic can append additional properties reported in the request telemetry.

Currently there is no easy way for a request handler to report additional properties other than the ones universally added, as success state, time spent etc. 

Returning additional fields in the return value might compromise the current semantic of the return value of each request. So I introduced an additional parameter to receive the additional telemetry properties.

This collector can also be used in the handlers for the notifications as needed.

### DIFF
--- a/src/chrome/chromeDebugSession.ts
+++ b/src/chrome/chromeDebugSession.ts
@@ -13,7 +13,7 @@ import { BaseSourceMapTransformer } from '../transformers/baseSourceMapTransform
 import { LineColTransformer } from '../transformers/lineNumberTransformer';
 
 import { IDebugAdapter } from '../debugAdapterInterfaces';
-import { telemetry, ExceptionType, IExecutionResultTelemetryProperties } from '../telemetry';
+import { telemetry, ExceptionType, IExecutionResultTelemetryProperties, TelemetryPropertyCollector } from '../telemetry';
 import * as utils from '../utils';
 import { ExecutionTimingsReporter, StepProgressEventsEmitter, IObservableEvents, IStepStartedEventsEmitter, IFinishedStartingUpEventsEmitter } from '../executionTimingsReporter';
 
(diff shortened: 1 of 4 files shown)


src/debugAdapterInterfaces.d.ts
```diff
@@ -179,7 +181,7 @@ export interface IDebugAdapter {
     shutdown(): void;
 
     initialize(args: DebugProtocol.InitializeRequestArguments, requestSeq?: number): PromiseOrNot<DebugProtocol.Capabilities>;
-    launch(args: ILaunchRequestArgs, requestSeq?: number): PromiseOrNot<void>;
+    launch(args: ILaunchRequestArgs, telemetryPropertyCollector: ITelemetryPropertyCollector, requestSeq?: number): PromiseOrNot<void>;
```
changsi-an: After I get some consensus about this design. I can modify all the request methods as well.

src/chrome/chromeDebugSession.ts
```diff
@@ -124,8 +124,9 @@ export class ChromeDebugSession extends LoggingDebugSession implements IObservab
      * Overload dispatchRequest to the debug adapters' Promise-based methods instead of DebugSession's callback-based methods
      */
     protected dispatchRequest(request: DebugProtocol.Request): void {
+        const telemetryPropertyCollector = new TelemetryPropertyCollector();
```
digeff: What are the advantages of creating the TelemetryPropertyCollector here instead of receiving it from the reportTelemetry method?

changsi-an: So less methods need to know the contract of requestHandler methods.

changsi-an: I think they are equivalent.

digeff: What do you think would make the code simpler and easier to maintain?
A. Keeping as much of the telemetry implementation inside the reportTelemetry method, and having reportTelemetry create the telemetryPropertyCollector so that the business methods don't need to deal with details of the telemetry implementation?
B. Spreading the telemetry implementation between reportTelemetry and the business methods, and having each business method that uses telemetry having to create it's own telemetryPropertyCollector?

roblourens:
What extra properties will this be used for?

changsi-an:
@roblourens This will be highly re-used for some notification handlers. Say when a breakpoint is hit, we need to know whether is a real or pesudo(break-on-load) breakpoint, whether it's for JS or TS. Currently our telemetry facility can only measure a processing time, successfulness etc. It does not support adding customized properties per the business of each handler.

roblourens:
I don't understand, since this is hooked up for client requests, how would you use it for notifications like onPaused?

I think it makes sense for client requests but for cases where ChromeDebugAdapter fires the telemetry events itself, it could just pass in any extra properties to reportEvent that it needs to, right?

changsi-an:
You are right. We haven't implemented that for notifications. This PR is just an example showing how a TelemetryCollector can be used.

For OnPaused. since we already has a wrapper to measure processing time in a universal way like this:
````
            this.runAndMeasureProcessingTime('target/notification/onPaused', () => {
                return this.onPaused(params);
            });
```

The `runAndMeasureProcessingTime` method will create a telemetry entry to describe what/how the onPause handler does. It is true that we have can have onPause call another telemetry.reportEvent(), but that would become a new telemetry entry. We have to merge everything into the one emitted by runAndMeasureProcessingTime, so we don't have to join/correlate telemetry entries down the pipeline.

PR #316, opened 2018-03-23
Now the start-up timings events can include a failure reason

digeff:
When we detect that the user requests for an unreachable url, we'll put that in the reason.

This change will break the debug adapters until we update them to use the new parameter required for the new signature.
This is related to this chrome-debug PR: https://github.com/Microsoft/vscode-chrome-debug/pull/630


roblourens:
> This change will break the debug adapters until we update them to use the new parameter required for the new signature.

Good excuse to bump to 4.0.0 :)

PR #317, opened 2018-03-26
Drop some telemetry events of a homogeneous type, if they accumulated too many within the telemetry sending interval.

changsi-an:
For a high-frequency event like onScriptParsed, we have the potential of hitting a per property size limit (1kB). So the aggregated property value (which has the form of "[v1,v2,v3,v4]") will be truncated at 1KB and become "[v1,v2,v3,". This cannot be processed correctly in the telemetry pipeline and we lose the data of the entire payload. So we truncate the data before its being sent out, hoping that the remainder is representative enough and can be formatted correctly. This can at least salvage some part of the data instead of ruining the entire payload. 

The change also allows such truncation behavior to be configured on a per-event type basis.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -392,8 +392,13 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             });
         });
         this.chrome.Debugger.onResumed(() => this.onResumed());
+
+        const onScriptParsedEventName = 'target/notification/onScriptParsed';
+        this._batchTelemetryReporter.configureEvent(onScriptParsedEventName, {
+            batchCap: 50
+        });
         this.chrome.Debugger.onScriptParsed((params) => {
-            this.runAndMeasureProcessingTime('target/notification/onScriptParsed', () => {
+            this.runAndMeasureProcessingTime(onScriptParsedEventName, () => {
                 return this.onScriptParsed(params);
             });
         });
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -141,14 +141,21 @@ export class BatchTelemetryReporter {
         for (const eventName in this._eventBuckets) {
             let bucket = this._eventBuckets[eventName];
             const eventConfig = this._eventConfig[eventName];
+            let properties: {[propertyName: string]: string} = {};
             if (eventConfig) {
                 const diff = bucket.length - eventConfig.batchCap;
+                let dropped: number;
                 if (diff > 0) {
                     bucket = bucket.slice(0, eventConfig.batchCap);
                     logger.log(`Truncate ${diff} ${eventName} event entries.`);
+                    dropped = diff;
+                } else {
+                    dropped = 0;
                 }
+                properties['aggregated.dropped'] = dropped.toString();
             }
-            let properties = BatchTelemetryReporter.transfromBucketData(bucket);
+
+            Object.assign(properties, BatchTelemetryReporter.transfromBucketData(bucket));
             this._telemetryReporter.reportEvent(eventName, properties);
         }
 
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -3,7 +3,7 @@
  *--------------------------------------------------------*/
 
 import { DebugProtocol } from 'vscode-debugprotocol';
-import { OutputEvent } from 'vscode-debugadapter';
+import { OutputEvent, logger } from 'vscode-debugadapter';
 import { fillErrorDetails } from './utils';
 
 export type ExceptionType = 'uncaughtException' | 'unhandledRejection' | 'firstChance';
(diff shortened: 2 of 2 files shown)


changsi-an:
@digeff  @dhanvikapila PTAL

src/telemetry.ts
```diff
@@ -129,7 +139,15 @@ export class BatchTelemetryReporter {
...
     private send(): void {
         for (const eventName in this._eventBuckets) {
-            const bucket = this._eventBuckets[eventName];
+            let bucket = this._eventBuckets[eventName];
+            const eventConfig = this._eventConfig[eventName];
+            if (eventConfig) {
+                const diff = bucket.length - eventConfig.batchCap;
+                if (diff > 0) {
+                    bucket = bucket.slice(0, eventConfig.batchCap);
+                    logger.log(`Truncate ${diff} ${eventName} event entries.`);
```
digeff: If we ever drop events we'll probably want to know how many total events we got, how many were successful, and the total/average time they took to execute

src/chrome/chromeDebugAdapter.ts
```diff
@@ -392,8 +392,13 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             });
         });
         this.chrome.Debugger.onResumed(() => this.onResumed());
+
+        const onScriptParsedEventName = 'target/notification/onScriptParsed';
+        this._batchTelemetryReporter.configureEvent(onScriptParsedEventName, {
+            batchCap: 50
```
digeff: If the truncation is based on property size of 1KB, would it make more sense to add all the events we can fit in under 1 KB, and then drop events or do something smart if we detect that a property is going over 1KB if we keep adding events?

changsi-an: We should do a trade-off between efficiency and quality here. We are talking about CPU cycles on customers machine here. I don't want to spend cycles on customer machine do complicated calculations or producing obscure code and reducing maintainability. And the end result is that we merely manage to salvage <10% more data. Does it really make a difference? I doubt that.

src/telemetry.ts
changsi-an: It can be done. But it's out of the scope of this PR. Let's figure out and reach to an agreement on what is needed to be done on the client side and understand the issues if we don't have this data, then decide if it's absolutely necessary to take that into regard.

Why do you think the dropped data needed to be taken into regard? How would it affect the insight we are trying to deduce?

dhanvikapila: I think at the very least, we should have some counter of how many we are dropping.  Averages and other aggregation we can add later.

src/chrome/chromeDebugAdapter.ts
dhanvikapila: Isn't it better to have a limit of say 100 and drop everything after that? Any reason we think that sample size will not be good sample set to get the distribution?

src/telemetry.ts
digeff: What I'm hoping to avoid is having some notifications with errors, or some notifications that are taking a very long time to get processed, and not getting that information because we drop those events.

src/telemetry.ts
```diff
@@ -129,7 +139,15 @@ export class BatchTelemetryReporter {
 
     private send(): void {
         for (const eventName in this._eventBuckets) {
-            const bucket = this._eventBuckets[eventName];
+            let bucket = this._eventBuckets[eventName];
+            const eventConfig = this._eventConfig[eventName];
+            if (eventConfig) {
+                const diff = bucket.length - eventConfig.batchCap;
+                if (diff > 0) {
+                    bucket = bucket.slice(0, eventConfig.batchCap);
```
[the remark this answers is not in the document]
changsi-an: No. Why do you think so?

src/telemetry.ts
```diff
@@ -97,15 +97,25 @@ export class NullTelemetryReporter implements ITelemetryReporter {
 
 export const DefaultTelemetryIntervalInMilliseconds = 10000;
 
+export interface IBatchTelemetryEventConfiguration {
+    batchCap: number;
```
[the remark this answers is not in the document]
changsi-an: It's at where you made the comment "Isn't it better to have a limit of say 100 and drop everything after that? Any reason we think that sample size will not be good sample set to get the distribution?"

src/chrome/chromeDebugAdapter.ts
```diff
@@ -392,8 +392,13 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
+        });
         this.chrome.Debugger.onScriptParsed((params) => {
-            this.runAndMeasureProcessingTime('target/notification/onScriptParsed', () => {
+            this.runAndMeasureProcessingTime(onScriptParsedEventName, () => {
```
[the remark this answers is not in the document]
changsi-an: No. The string literal is extracted as a constant and put above.

src/chrome/chromeDebugAdapter.ts
changsi-an: We have data like

```
VS.BPT.Diagnostics.ChromeDebugAapter.aggregated.startTime	["1522092607233","1522092607391"]
```
This is a good example of a maximum length each data element can get, which is 15 bytes on average. 100 of them will reach 1500 KB, way over 1KB.

```
Will we be able to process the information if the property with 100 values gets cut because it exceeds 1 KB?
```

No, if a property carries 100 values and gets cut, it become invalid. There is no reliable way to recover these 100 values. Potentially we can recover a part of it but it's tricky, merely adding complexity to the downstream processing.

changsi-an: If the point of discussion is how big the number should be, I would recommend discussing it offline. We can easily change it to whatever value. It does not have to block other code change in this PR.

src/telemetry.ts
changsi-an: I can add a counter for how many we dropped.

@digeff From a statistical point of view, I incline to assume that the remaining part follows the same distribution as the rest. In other words, notification with error and long processing has equal opportunity to appear in the retained portion.

roblourens:
Do @digeff and @dhanvikapila sign off? Is this ready to go?

digeff:
@roblourens Chance is currently OOF. Is it okay if we hold off this PR for a few days?

roblourens:
Ure

roblourens:
@digeff @changsi-an Is this PR still relevant?

dhanvikapila:
No. Go ahead and discard it.


Thanks.

- Dhanvi

________________________________

PR #318, opened 2018-03-29
Clearing the context immediately to avoid sending multiple remove eve…

rakatyal:
…nts in case of quick refreshes

When we have a case of multiple refreshes in the browser in a quick succession, sometimes we have a case where we don't reach the point of clearing the context and, in turn _scriptById, while sending out 'remove' events from the previous refresh. 
In this scenario, the subsequent refresh gets the _scriptsById of the previous state and we end up sending multiple remove events for the same script causing PineZorro to throw an error.

We should clear the context as soon as we populate the cachedScriptParsedEvents array.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -533,13 +533,12 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
      */
     protected onExecutionContextsCleared(): Promise<void> {
         const cachedScriptParsedEvents = Array.from(this._scriptsById.values());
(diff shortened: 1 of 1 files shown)


roblourens:
@rakatyal I won't merge it this week because we are about to ship vscode. Do you need it published right away? If so, I can merge it to another branch and publish. Just let me know...

rakatyal:
@roblourens: Okay. So when can we merge this earliest and get it published via master?

roblourens:
End of this week?

rakatyal:
@roblourens: Can we merge and publish this today?

PR #320, opened 2018-04-04
Update noice to latest version

digeff:
When the latest version fails it rejects the promises with actual errors, so we'll get stack traces on telemetry if something goes wrong.


roblourens:
Thanks for getting that in

PR #323, opened 2018-04-09
Now we log the unfiltered targets

digeff:
Now we log the unfiltered targets


roblourens:
I just fixed it a different way, I don't want to have `targets` and `filteredTargets` and `unfilteredTargets`

PR #329, opened 2018-05-17
Now we keep a collection to determine if a script needs to be added, removed or changed on the client

digeff:
Now we keep a collection to determine if a loaded script needs to be added, removed or changed on the client.

This fixes an issue on Edge where we are trying to remove scripts that are not there, and trying to add scripts that already are there.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -156,6 +156,8 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
     public readonly events: StepProgressEventsEmitter;
 
+    private _loadedSourcesByScriptId = new Map<Crdp.Runtime.ScriptId, CrdpScript>();
+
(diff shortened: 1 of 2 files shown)


digeff:
@rakatyal Can you take a look please?

roblourens:
> This fixes an issue on Edge where we are trying to remove scripts that are not there, and trying to add scripts that already are there.

Why does that happen?

digeff:
We have a race condition between: onExecutionContextsCleared() and onScriptParsed(). When a script gets added to this._earlyScripts we can try to remove it before we add it.

In Edge this happens often at the moment because of a bug when we navigate to a new page we get a onScriptParsed() for the new page before onExecutionContextsCleared() is called.

I think this can probably also happen on Chrome, but I haven't seen it happen yet.

auchenberg:
Question: If this only happens in Edge, why isn't this fixed in the platform?

digeff:
Edge platform is fixing their bug too. I'm fixing this here because we want to get this working as soon as possible, and also because I think that this can potentially happen on Chrome too.

roblourens:
I don't think this happens in Chrome, I think this is duplicating information that the browser should keep track of. I'm ok taking it as a workaround for the bug as long it we mark it as such. Will look at it on Monday.

PR #332, opened 2018-06-04
Fix break-on-load to make breakpoints on attach work

digeff:
* Move this._breakOnLoadHelper to common args so break on load can be used for both launch and attach
* Update the logic of getUrlRegexForBreakOnLoad to handle case insensitive paths
  * Pass caseSensitivePaths around which is an argument of the case insensitive path logic

This as final bug-fixes. At this moment we are using this change as a temporary-workaround to fix the NTVS unit tests


digeff:
@rakatyal PTAL

rakatyal:
@roblourens: Yes I believe this is still valid but we are not planning to ship this at the moment. I will review this soon.

digeff:
@roblourens @rakatyal We are planning on using this PR for VS 16. Could you please review it now?

roblourens:
@rakatyal are you going to review this?

rakatyal:
@roblourens: Yes I will do this today. Apologies for missing this earlier.

test/chrome/chromeDebugAdapter.test.ts
```diff
@@ -860,6 +859,18 @@ suite('ChromeDebugAdapter', () => {
         });
     });
 
+    suite('break-on-load', () => {
+        test('is active when the parameter is specified and we are launching', async () => {
+            await chromeDebugAdapter.launch({breakOnLoadStrategy: 'regex'});
+            assert(chromeDebugAdapter.breakOnLoadActive, 'Break on load should we active if we pass the proper parameter and we are attaching');
```
rakatyal: should "be" active

test/chrome/chromeDebugAdapter.test.ts
```diff
@@ -860,6 +859,18 @@ suite('ChromeDebugAdapter', () => {
...
+        });
+
+        test('is active when the parameter is specified and we are attaching', async () => {
+            await chromeDebugAdapter.attach({breakOnLoadStrategy: 'regex', port: ATTACH_SUCCESS_PORT});
+            assert(chromeDebugAdapter.breakOnLoadActive, 'Break on load should we active if we pass the proper parameter and we are attaching');
```
rakatyal: should "be" active

Issue #334, opened 2018-06-16
Debugger crashes with TypeError: Cannot read property 'url' of undefined


jhnns:
When I'm trying to debug my code with VS Code, the debugger fails and logs the following stack trace to the debug console:

```
******** Unhandled error in debug adapter: TypeError: Cannot read property 'url' of undefined
    at isLogpointMessage (/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/ms-vscode.node-debug2/node_modules/vscode-chrome-debug-core/out/src/chrome/internalSourceBreakpoint.js:28:54)
    at Object.stackTraceWithoutLogpointFrame (/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/ms-vscode.node-debug2/node_modules/vscode-chrome-debug-core/out/src/chrome/internalSourceBreakpoint.js:31:9)
    at NodeDebugAdapter.onConsoleAPICalled (/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/ms-vscode.node-debug2/node_modules/vscode-chrome-debug-core/out/src/chrome/chromeDebugAdapter.js:955:50)
    at NodeDebugAdapter.onConsoleAPICalled (/Applications/Visual Studio Code.app/Contents/Resources/app/extensions/ms-vscode.node-debug2/out/src/nodeDebugAdapter.js:382:19)
    ...
```

The problem comes from [this line](https://github.com/Microsoft/vscode-chrome-debug-core/blob/65da9157b7a3f706210fffced3b65c535ce96eb9/src/chrome/internalSourceBreakpoint.ts#L33) where  `callFrames` is expected to be an array with at least one item:

```js
function isLogpointMessage(m: Crdp.Runtime.ConsoleAPICalledEvent): boolean {
    return m.stackTrace && m.stackTrace.callFrames[0].url === InternalSourceBreakpoint.LOGPOINT_URL;
}
```

At the time when the error happens, `m` looks like:

```js
{
    "type": "error",
    "args": [
        {
            "type": "object",
            "subtype": "error",
            "className": "NodeError",
            "description":
[...]

PR #335, opened 2018-06-16
Fix unsafe array access in internalSourceBreakpoint

jhnns:
The `stackTrace.callFrames` property of the `Protocol.Runtime.ConsoleAPICalledEvent` can be an empty array in certain conditions.

See https://github.com/Microsoft/vscode-chrome-debug-core/issues/334


roblourens:
Thanks for the PR!

PR #336, opened 2018-06-20
Fix possible race condition for new and changed scripts

digeff:
We saw on one of our automated tests that we sent a "changed" event for a script before the "new" event. This fixes one potential cause.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -21,7 +21,7 @@
             "request": "launch",
             "protocol": "inspector",
             "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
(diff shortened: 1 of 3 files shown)


roblourens:
I don't understand what this is fixing, nothing can happen between the top of that method and the original place where `scriptToSource` was called, right?

digeff:
If we get the same script twice, and we put the reason 'new' first, and 'changed' later, and then we call this.scriptToSource() maybe the 'new' event will get blocked for longer than the 'changed' event', and we'll end up sending the 'changed' event before the 'new' event.

We've seen that behavior happen in our tests, and VS fails because of the unexpected order. I don't know for sure if this is the cause, but it's one of the potential places that could be causing that.

roblourens:
Blocked by what? In javascript, the thread won't be interrupted in sync code. There are no awaits between lines 863 and 882 so I don't see how that would happen. Does this fix the problem in your tests?

PR #337, opened 2018-06-21
Persist the client ID when we reset the breakpoints in source mapped …

rakatyal:
…files

When we clear and reset the  breakpoints in the mapped files, we don't persist the client IDs we created for them and sent to PineZorro. This specially causes a problem in Edge which returns a new breakpoint ID when we make the setBreakpointByUrl call again during resetting. Then we end up sending a 'changed' event to PZ for this breakpoint ID which causes an error since we didn't tell the client about it.

This doesn't cause an issue with Chrome currently because Chrome uses a pattern of script_url, line and column number to create the breakpoint ID which remains the same on resetting the breakpoint again. 

This change persists the client IDs during the reset process and makes sure it gets mapped to the original one.

### DIFF
--- a/src/transformers/baseSourceMapTransformer.ts
+++ b/src/transformers/baseSourceMapTransformer.ts
@@ -40,6 +40,7 @@ export class BaseSourceMapTransformer {
     private _requestSeqToSetBreakpointsArgs: Map<number, ISavedSetBreakpointsArgs>;
     private _allRuntimeScriptPaths: Set<string>;
     private _authoredPathsToMappedBPs: Map<string, DebugProtocol.SourceBreakpoint[]>;
+    private _authoredPathsToClientBreakpointIds: Map<string, number[]>;
 
     protected _preLoad = Promise.resolve();
     private _processingNewSourceMap: Promise<any> = Promise.resolve();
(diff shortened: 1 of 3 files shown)


src/transformers/baseSourceMapTransformer.ts
```diff
@@ -120,6 +122,11 @@ export class BaseSourceMapTransformer {
 
                 this._authoredPathsToMappedBPs.set(argsPath, args.breakpoints);
 
+                // Store the client breakpoint Ids for the mapped BPs as well
+                if (ids) {
+                    this._authoredPathsToClientBreakpointIds.set(argsPath, ids);
```
roblourens: Does this need to be in a map on the instance? Since they aren't referenced outside of this method.

src/transformers/baseSourceMapTransformer.ts
```diff
@@ -40,6 +40,7 @@ export class BaseSourceMapTransformer {
     private _requestSeqToSetBreakpointsArgs: Map<number, ISavedSetBreakpointsArgs>;
     private _allRuntimeScriptPaths: Set<string>;
     private _authoredPathsToMappedBPs: Map<string, DebugProtocol.SourceBreakpoint[]>;
+    private _authoredPathsToClientBreakpointIds: Map<string, number[]>;
```
roblourens: I hate that we have to keep track of the breakpoint IDs in here. But we're already keeping track of the breakpoints anyway. It's not great. Let me think about this a little more.

rakatyal: Okay let me know!

rakatyal:
@digeff , @mrcrane FYI.

src/transformers/baseSourceMapTransformer.ts
rakatyal: @roblourens: Did you get a chance to think about this yet?

roblourens:
@rakatyal please check the unit tests. VSTS isn't building PRs for some reason.

PR #339, opened 2018-07-03
Fixing unit tests

rakatyal:


rakatyal:
Ahh looks like you already fixed it. Thanks!

PR #341, opened 2018-07-16
Canonicalize paths to lowercase in Windows

digeff:
Canonicalize paths to lowercase in Windows

In some cases the webroot that VS sends doesn't match the case of the file paths that VS sends for the breakpoints: ("c:\users\username\source\repos\WebApplication77\WebApplication77" vs "c:\Users\username\source\repos\WebApplication77\WebApplication77\Scripts\bootstrap.js"). That results in breakpoints not being hit in those cases.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -122,8 +122,19 @@ export function canonicalizeUrl(urlOrPath: string): string {
 
     urlOrPath = stripTrailingSlash(urlOrPath);
     urlOrPath = fixDriveLetterAndSlashes(urlOrPath);
+    urlOrPath = normalizeIfFSIsCaseInsensitive(urlOrPath);
 
     return urlOrPath;
+}
+
+function normalizeIfFSIsCaseInsensitive(urlOrPath: string): string {
+    return (getPlatform() === Platform.Windows && isFilePath(urlOrPath))
+        ? urlOrPath.toLowerCase()
+        : urlOrPath;
+}
+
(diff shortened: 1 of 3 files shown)


roblourens:
This isn't handled by the code we already have for this?  https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/chrome/chromeDebugAdapter.ts#L2873

And this needs to be VS-only like the other code.

digeff:
I can reproduce this issue with the latest version of the debug adapter, so the existing code is not handling this case.
I can reproduce this on VS Code by hardcoding the "webRoot" parameter manually and using the wrong case, so I think we should make this fix for both VS and VS Code.

roblourens:
We already have a VS-specific fix, we need to find the missing scenario and use that same fix. If someone uses the wrong case in VS Code, then they should fix that.

digeff:
By same fix: Do you mean enabling this code only if _caseSensitivePaths  is set to true?

roblourens:
Yeah, but you shouldn't have to check that inside this utils method. The existing fix should handle all paths coming and going from the adapter.

digeff:
What is your concern with putting this logic inside the utils method?

My concern with trying to handle all paths coming and going from the adapter is that if we miss one, then we have a bug. It seems more reliable to me to put this logic on the util method, so even if we add new ways (or miss old ones) for paths to come and go from the debug adapter, everything will work fine.

PR #342, opened 2018-07-18
Canonicalize win32 paths to lowercase

digeff:
Canonicalize win32 paths to lowercase

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -103,7 +103,7 @@ export function retryAsync(fn: () => Promise<any>, timeoutMs: number, intervalDe
     return tryUntilTimeout();
 }
 
-var caseSensitivePaths: boolean = true;
+let caseSensitivePaths = true;
 export function setCaseSensitivePaths(useCaseSensitivePaths: boolean) {
     caseSensitivePaths = useCaseSensitivePaths;
 }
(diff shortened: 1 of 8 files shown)


src/utils.ts
```diff
@@ -122,10 +127,23 @@ export function canonicalizeUrl(urlOrPath: string): string {
...
     return urlOrPath;
 }
 
+function normalizeIfFSIsCaseInsensitive(urlOrPath: string): string {
+    return isFilePath(urlOrPath)
+        ? urlOrPath.toLowerCase()
+        : urlOrPath;
+}
+
+export function isFilePath(candidate: string): boolean {
```
mrcrane: This will only match Windows-style paths. If it won't match unix-style paths, can you give the method a more specific name? Does it need to be exported?

digeff: Changed to isWindowsFilePath

auchenberg:
@digeff Why? What's the problem being solved here? Please provide some context for the reviewers here. Should this happen on all platforms? etc?

digeff:
@auchenberg This is Windows only
We get paths from different places, and they can have different cases even if they refer to the same file. We need to make sure that the code doesn't consider the case when comparing file paths.
In VS we get the webRoot from the ProjectSystem, the file paths from the Debugger, and other file paths from Chrome or Node. In node the case depends on the case used on the require('myFile') statement. There is no easy way to make sure that all those components use the same case, so the solution seems to be to make sure that the comparison is always case insensitive.

roblourens:
@digeff Could you write a test that captures the case that was missed, and prompted this PR? I think you said it was for the webroot?

PR #345, opened 2018-08-01
Logging and a small fix while investigating a test issue

digeff:
Added some logging while investigating a test issue.

Also removed the utils.fixDriveLetter calls. Given that we call utils.canonicalizeUrl in the previous line, that should make the utils.fixDriveLetter unneccesary

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -845,7 +845,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
                     await this.resolvePendingBreakpoint(pendingBP);
                     this._pendingBreakpointsByUrl.delete(source);
                 } else {
-                    const sourceFileName = path.basename(source);
+                    const sourceFileName = path.basename(source).toLowerCase();
                     if (Array.from(this._pendingBreakpointsByUrl.keys()).find(key => key.toLowerCase().indexOf(sourceFileName) > -1)) {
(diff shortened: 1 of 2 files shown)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -839,10 +839,14 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
             const resolvePendingBPs = async (source: string) => {
                 source = source && utils.canonicalizeUrl(source);
-                const pendingBP = this._pendingBreakpointsByUrl.get(utils.fixDriveLetter(source)) || this._pendingBreakpointsByUrl.get(utils.fixDriveLetter(source, true));
+                logger.log(`OnScriptParsed.resolvePendingBPs: pendingBreakpointsByUrl = ${JSON.stringify([...this._pendingBreakpointsByUrl])} source = ${source}`);
```
roblourens: I don't think you need to copy `this._pendingBreakpointsByUrl` (just JSON.stringify(this. ...)

roblourens: Oh it's a map, I see. Do you need to log the whole thing, and the individual bp below?

digeff: It was useful while I was debugging the test. I'm not sure what logging we'll need for the next issue.

roblourens: This message will be printed for every loaded script whether or not there are pending BPs, right? Can you cut it down to just the necessary stuff to avoid too much log spam?

digeff: You are right, so much spam will be annoying. What do you think of this filter?

roblourens:
This broke tests, please fix

PR #347, opened 2018-08-08
Add workaround for the ASP.NET webRoot being incorrectly configured

digeff:
We currently can't hit breakpoints in VS with ASP.NET Core 2.1 for Angular.
The challenge is that depending on the runtime configuration of angular the webRoot folder might be : <workspace>\ClientApp or it might be <workspace>\wwwRoot and we don't have any way of figuring out which case it is before we launch the debugger.

For that reason we are implementing this fallback as a temporary workaround so we'll try the normal webRoot and if that doesn't work we'll try <workspace>\ClientApp to see if that works

P.S: After we all agree on how to implement this fix I'll add telemetry and a test for the final code

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -33,6 +33,7 @@ import { BaseSourceMapTransformer } from '../transformers/baseSourceMapTransform
 import { EagerSourceMapTransformer } from '../transformers/eagerSourceMapTransformer';
 import { FallbackToClientPathTransformer } from '../transformers/fallbackToClientPathTransformer';
 import { BreakOnLoadHelper } from './breakOnLoadHelper';
+import * as sourceMapUtils from '../sourceMaps/sourceMapUtils';
 
 import * as path from 'path';
 
@@ -1336,6 +1337,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         }
     */
     public disconnect(args: DebugProtocol.DisconnectArguments): void {
+        telemetry.reportEvent('FullSessionStatistics/SourceMaps/Overrides', { aspNetClientAppFallbackCount: sourceMapUtils.getAspNetFallbackCount() });
         this.shutdown();
         this.terminateSession('Got disconnect request', args);
     }
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -161,6 +161,8 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
     private _loadedSourcesByScriptId = new Map<Crdp.Runtime.ScriptId, CrdpScript>();
 
+    private _isVSClient: boolean;
+
     public constructor({ chromeConnection, lineColTransformer, sourceMapTransformer, pathTransformer, targetFilter, enableSourceMapCaching }: IChromeDebugAdapterOpts,
         session: ChromeDebugSession) {
         telemetry.setupEventHandler(e => session.sendEvent(e));
@@ -235,7 +237,9 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             this._pathTransformer = new FallbackToClientPathTransformer(this._session);
         }
 
-        utils.setCaseSensitivePaths(args.clientID !== 'visualstudio');
+        this._isVSClient = args.clientID === 'visualstudio';
+        utils.setCaseSensitivePaths(!this._isVSClient);
+        this._sourceMapTransformer.isVSClient = this._isVSClient;
 
         if (args.pathFormat !== 'path') {
             throw errors.pathFormat();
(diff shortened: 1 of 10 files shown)


roblourens:
Can't you check for which folder exists before launching the debugger?

I really don't like checks like this, it makes the sourcemap resolving process hard to follow. And this is not VS-specific or ASP.net specific, it could affect lots of users.

digeff:
For the ASP.NET 2.0 the ClientApp folder exists and we need to use /wwwroot as the webRoot.
For the ASP.NET 2.1 the ClientApp folder also exists, and we need to use /ClientApp as the webRoot
Knowing that the ClientApp folder is there doesn't tell us which webRoot parameter to use.

I can add a useVSASPNETFix: bool parameter and pass it from ChromeDebugAdapter so that this logic is only enabled for VS.

Can you think of a different way to make this work that doesn't make the sourcemap resolving process more hard to follow?

roblourens:
So both folders always exist in both cases, but you don't know which one the source files are under?

How does this work in normal ASP.net scenarios, how does the system know where the users files are?

roblourens:
Can you look at the contents of the wwroot/ClientApp folder to tell where the actual project is? Is there something obvious like a package.json or index.js or a 'src' folder to find?

digeff:
We can do some checks that will work most of the times. We don't know how to do a check that will work all the time.

If we implement the check that works "most of the time" we'll be breaking the debugger experience for some cases where it currently works. We prefer this fallback approach because we won't break any case that was working before this change.

roblourens:
Does that mean that files in a single project are split between the folders?

Sorry but I'm just trying to make sure we've covered all the bases. If the problem is, an Angular project is sometimes in folder A and sometimes in folder B, that seems easy to detect. In that case I don't see why we need this dynamic per-file checking which only covers one specific case and will never go away.

digeff:
The angular project is always in folder ClientApp/. The difference is that sometimes the .js files are generated and put in wwwRoot, and sometimes they are kept in memory and they are not on wwwRoot.

The reason that check things like "are .js files present" on wwwRoot doesn't work for all cases is because the user can have other libraries that put files on wwwRoot and angular with files on memory at the same time.

roblourens:
Actually in this case you are looking for the .ts files not the .js files right?

roblourens:
Or whatever the source files are.

roblourens:
Maybe I should come over tomorrow and look at this because if the source files are always in the same place, then sourceMapPathOverrides can just point there, you don't necessarily need to know where the generated scripts are.

digeff:
I still need to add telemetry and logging to this change

src/sourceMaps/sourceMapUtils.ts
```diff
@@ -64,11 +64,16 @@ export function getComputedSourceRoot(sourceRoot: string, generatedPath: string,
     return absSourceRoot;
 }
 
+let aspNetFallbackCount = 0;
```
rakatyal: What is the significance of this variable?

digeff: We use it to count how many times the "fallback" bevahior is happening. This will also tell us how many users are using this fix. We send it as telemetry.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1332,6 +1337,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         }
     */
     public disconnect(args: DebugProtocol.DisconnectArguments): void {
+        telemetry.reportEvent('FullSessionStatistics/SourceMaps/Overrides', { aspNetClientAppFallbackCount: sourceMapUtils.getAspNetFallbackCount() });
```
rakatyal: I don't recall it but did Michael find issues with telemetry reporting after we issue the disconnect request? Can you confirm we will get this at all instances?

digeff: I tested this a few times on my machine and the telemetry did appear on the telemetry test tool.

src/sourceMaps/sourceMapUtils.ts
rakatyal: Wouldn't this count just reflect on the number of files in the user's project? What I implied was if storing just yes/no if this feature was needed be enough for telemetry purposes.

src/chrome/chromeDebugAdapter.ts
rakatyal: Cool thanks!

src/sourceMaps/sourceMapUtils.ts
digeff: It's the same effort to send yes vs. no than a number.
It could be useful to know this number a certain point (maybe we see that the debugger is very slow when this number gets bigger, so there might be a performance issue in this fix)

rakatyal:
@digeff: Looks like your new tests are failing.

digeff:
yeah :(
working on that...

roblourens:
Summary of our offline conversation - VS needs to be able to debug any template configuration out of the box but the broken template is already shipped, and VS doesn't know which template was used to generate the project. The sourcemaps are different between the two projects so we need to apply different overrides. There isn't a safe way to look at things on disk before launching to decide which config to use. And the most important thing is to avoid breaking anything that is currently working.

PR #348, opened 2018-08-29
Add property to get version of chrome debug protocol

digeff:
Add property to get version of chrome debug protocol.

We need to do some version-checking on edge-debug ( https://github.com/Microsoft/vscode-edge-debug2/pull/100 )

### DIFF
--- a/src/chrome/chromeConnection.ts
+++ b/src/chrome/chromeConnection.ts
@@ -9,7 +9,7 @@ import { StepProgressEventsEmitter, IObservableEvents, IStepStartedEventsEmitter
 import * as errors from '../errors';
 import * as utils from '../utils';
 import { logger } from 'vscode-debugadapter';
-import { ChromeTargetDiscovery } from './chromeTargetDiscoveryStrategy';
+import { ChromeTargetDiscovery, ChromeDebugProtocolVersion } from './chromeTargetDiscoveryStrategy';
 
 import { Client, LikeSocket } from 'noice-json-rpc';
 
@@ -31,6 +31,8 @@ export interface ITarget {
 
 export type ITargetFilter = (target: ITarget) => boolean;
 export interface ITargetDiscoveryStrategy {
+    version: Promise<ChromeDebugProtocolVersion>;
+
     getTarget(address: string, port: number, targetFilter?: ITargetFilter, targetUrl?: string): Promise<ITarget>;
     getAllTargets(address: string, port: number, targetFilter?: ITargetFilter, targetUrl?: string): Promise<ITarget[]>;
 }
@@ -173,5 +175,9 @@ export class ChromeConnection implements IObservableEvents<IStepStartedEventsEmi
 
     public onClose(handler: () => void): void {
         this._socket.on('close', handler);
+    }
+
+    public get version(): Promise<ChromeDebugProtocolVersion> {
+        return this._targetDiscoveryStrategy.version;
     }
 }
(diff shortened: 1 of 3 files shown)


roblourens:
What exactly does this do and why do you need to do this? Can Edge expose whatever it needs to do, some other way than version number checks?

digeff:
In RS4 both Debugger.msSetDebuggerPropertyValue and Runtime.callFunctionOn were available. Runtime.callFunctionOn only worked for some cases. Debugger.msSetDebuggerPropertyValue was created to handle the cases that Runtime.callFunctionOn didn't work on.

In RS5 edge finally supports Runtime.callFunctionOn  for all cases, so Debugger.msSetDebuggerPropertyValue  it's not needed any more.

In the future we'd like to use /json/protocol to see if an API is available or not, and then use the APIs that are available. In this case we cannot do that because Runtime.callFunctionOn was already available in RS4 which is version 0.1, so /json/protocol for both 0.1 (RS4) and 0.2 (RS5) will both show Runtime.callFunctionOn available and with the same signature. So we need to use the Edge Protocol Version 0.2 (RS5) to figure out if we need to use Runtime.callFunctionOn  for all cases or not.

auchenberg:
The name of this objects suggests this to be a string,  but it's really an object with helper functions.  `ChromeDebugProtocolVersion.version` also seems redundant. I would suggest renaming to something like `ProtocolDetails`, `ProtocolSchema` or maybe add this a generic metadata on the target, and move the helper functions into it's own module or inside the needed DA.

digeff:
@auchenberg I renamed the class to ProtocolSchema like you suggested.

Could you please give me more details about what do you mean by "maybe add this a generic metadata on the target" and what is the advantage that you see in "move the helper functions into it's own module or inside the needed DA"?

Thanks!

auchenberg:
@digeff My suggestion was to either rename or move the logic outside of the core, and into the DA where the version detection is needed.

src/chrome/chromeTargetDiscoveryStrategy.ts
```diff
@@ -14,10 +14,23 @@ import { ITargetDiscoveryStrategy, ITargetFilter, ITarget } from './chromeConnec
 import * as nls from 'vscode-nls';
 const localize = nls.loadMessageBundle();
 
+export class ProtocolSchema {
+    public static unkownVersion(): any {
```
roblourens: Small typo, unknown

roblourens: Return type, ProtocolSchema, and use `new ProtocolSchema` instead of `new this`, I didn't even know `this` would work that way in a static method

digeff: fixed

src/chrome/chromeConnection.ts
```diff
@@ -31,6 +31,8 @@ export interface ITarget {
 
 export type ITargetFilter = (target: ITarget) => boolean;
 export interface ITargetDiscoveryStrategy {
+    version: Promise<ProtocolSchema>;
```
roblourens: I don't think `version` belongs as a property on this. It's otherwise stateless, ie you call `getTarget`/`getAllTargets` and get info back. This property would be "the version of the last endpoint I accessed" which doesn't feel right. So I think it would be better for `getTarget/getAllTargets` to return a result object that contains the ITargets and also the version. Does that make sense?

src/chrome/chromeConnection.ts
```diff
@@ -27,6 +27,7 @@ export interface ITarget {
     type: string;
     url?: string;
     webSocketDebuggerUrl: string;
+    version: Promise<ProtocolSchema>;
```
roblourens: Do you think we should wait on getting the version so it can just a ProtocolSchema instead of a promise?

PR #350, opened 2018-09-05
Add public interface to set targetFilter on connection at runtime.

stristr:
Please see [`vscode-chrome-debug` #713](https://github.com/Microsoft/vscode-chrome-debug/pull/713) for original (rejected) proposal for this feature.

I wanted to emphasize backward compatibility when designing the API, so I opted to leave `targetFilter` alone (it's not broken and might be useful for other extensions of `vscode-chrome-debug-core`). Instead, where necessary, I:
 - Added `targetTypes` as the last optional argument at the tail end of existing chains of optional arguments.
 - Exposed `setTargetTypes(targetTypes: string[])` as a public method on `ChromeConnection`. This felt like a good alternative to rewiring the whole mechanism by which `ChromeConnection` receives `targetFilter` for consistency. (Note: I would have used an ES6 setter here, but apparently TypeMoq has some trouble with setters in strict mode.)

A corresponding `vscode-chrome-debug` pull request is blocked by this PR, but already done/tested in my fork [here](https://github.com/stristr/vscode-chrome-debug/commit/a97bb09e6a22936ac1ea91541966dfc6ee67cc5f). If this checks out, would you please let me know the instructions for contributing sibling PRs for projects that use `vscode-chrome-debug-core`? Despite the backward-compatible API, other packages that mock `ChromeConnection` in strict mode will probably need something like:

```
        mockChromeConnection
            .setup(x => x.setTargetTypes(It.isValue(undefined)))
            .verifiable(Times.atLeast(0));
```

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -400,6 +400,11 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         args.sourceMaps = typeof args.sourceMaps === 'undefined' || args.sourceMaps;
 
         this._smartStepEnabled = this._launchAttachArgs.smartStep;
+
+        // Use hasOwnProperty to explicitly permit setting a falsy targetFilter.
+        if (args.hasOwnProperty('targetFilter')) {
+            this._chromeConnection.setTargetFilter(args.targetFilter);
+        }
     }
 
     public shutdown(): void {
--- a/src/chrome/chromeConnection.ts
+++ b/src/chrome/chromeConnection.ts
@@ -115,6 +115,10 @@ export class ChromeConnection implements IObservableEvents<IStepStartedEventsEmi
         return this._attachedTarget;
     }
 
+    public setTargetFilter(targetFilter?: ITargetFilter) {
+        this._targetFilter = targetFilter;
+    }
+
     /**
      * Attach the websocket to the first available tab in the chrome instance with the given remote debugging port number.
      */
--- a/src/debugAdapterInterfaces.d.ts
+++ b/src/debugAdapterInterfaces.d.ts
@@ -10,6 +10,7 @@ import { DebugProtocol } from 'vscode-debugprotocol';
 import { Protocol as Crdp } from 'devtools-protocol';
 import { ITelemetryPropertyCollector } from './telemetry';
 import { IStringDictionary } from './utils';
+import { ITargetFilter } from './chrome/chromeConnection';
 
 export type ISourceMapPathOverrides = IStringDictionary<string>;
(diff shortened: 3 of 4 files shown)


stristr:
(Another thing I wasn't sure how to do: contribute 13 translations documenting the new property in the `package.nls.*.json` files 😄…)

roblourens:
Thanks! I will review this soon...

stristr:
Thanks :smile:

BTW, I checked out all the dependent packages listed in this project's README and determined only `vscode-chrome-debug` will need changes. Others are deprecated, track a much older version of this dependency, or lack a test that would break due to the reasons listed in the PR description.

roblourens:
I agree that breaking other packages is probably not a big concern, thanks for doing that research.

I wonder whether we need ChromeTargetDiscoveryStrategy to know about these targetTypes, when it already has targetFilter which is basically just used by chrome-debug to accomplish the same thing. Is it possible to just pass in a targetFilter which filters to the correct target types? I can't remember whether I told you to do this one way or another last time around.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -400,6 +400,10 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         args.sourceMaps = typeof args.sourceMaps === 'undefined' || args.sourceMaps;
 
         this._smartStepEnabled = this._launchAttachArgs.smartStep;
+
+        // Filter down to a set of requested target types. Note that `targetTypes` might be `undefined` here, which is expected;
+        // the result will be disabling target type filtering unless explicitly requested by the consumer.
+        this._chromeConnection.setTargetTypes(args.targetTypes);
```
stristr: Yeah, I think we could pull it off without much friction…we'd just need to expose a public method like this one, so that ChromeConnection, in addition to possibly receiving `targetFilter` at construction time, also can receive target filter at args-time. This block would roughly change to:

```
if (args.targetFilter) {
  this._chromeConnection.setTargetFilter(args.targetFilter);
}
```

You want me to go with that approach instead?

stristr: (cc: @roblourens)

roblourens: Yeah, I think that makes sense.

stristr:
Okay @roblourens, done. PR title is updated to reflect this change, as is my fork of `vscode-chrome-debug` to use this API instead ([link](https://github.com/stristr/vscode-chrome-debug/commit/4c59042cf1331fea3fa4bbef86824df7ad90ecb5)).

roblourens:
Thanks!

PR #351, opened 2018-09-06
Fix bug in protocol schema

digeff:
Fix bug in protocol schema

I was doing needed_for_feature >= os
I should've been doing os >= needed_for_feature

### DIFF
--- a/src/chrome/chromeTargetDiscoveryStrategy.ts
+++ b/src/chrome/chromeTargetDiscoveryStrategy.ts
@@ -22,7 +22,7 @@ export class ProtocolSchema {
     constructor(private _major: number, private _minor: number) {}
 
     public isAtLeastVersion(major: number, minor: number): boolean {
-        return major > this._major || (major === this._major && minor >= this._minor);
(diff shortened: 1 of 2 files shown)


src/chrome/chromeTargetDiscoveryStrategy.ts
```diff
@@ -22,7 +22,7 @@ export class ProtocolSchema {
     constructor(private _major: number, private _minor: number) {}
 
     public isAtLeastVersion(major: number, minor: number): boolean {
-        return major > this._major || (major === this._major && minor >= this._minor);
+        return this._major > major || (this._major === major && this._minor >= minor);
```
dhanvikapila: Can `this._major` or `this._minor` be `undefined`/`null`?

digeff: It can't be undefined/null. It can be NaN if the format of /json/version is not what we expected.

In that case 2 > NaN etc... will all return false and behave the way we want it to...

PR #356, opened 2018-09-18
Shennie issue68on edge

Shenniey:
- exporting TargetVersions class so that vscode-edge-debug-adapter can properly load messages
- putting a check so that properties without display values show up as "not available", as opposed to "" (in reference to this issue: https://github.com/Microsoft/vscode-edge-debug2/issues/68)

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -2487,9 +2487,9 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
         // Convert to a Variable object then just copy the relevant fields off
         const variable = await this.remoteObjectToVariable(args.expression, evalResponse.result, /*parentEvaluateName=*/undefined, /*stringify=*/undefined, <VariableContext>args.context);
-        if (evalResponse.exceptionDetails) {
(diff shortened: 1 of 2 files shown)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -2487,9 +2487,9 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
 
         // Convert to a Variable object then just copy the relevant fields off
         const variable = await this.remoteObjectToVariable(args.expression, evalResponse.result, /*parentEvaluateName=*/undefined, /*stringify=*/undefined, <VariableContext>args.context);
-        if (evalResponse.exceptionDetails) {
+        if (evalResponse.exceptionDetails || !variable.value) {
             let resultValue = variable.value;
-            if (resultValue && (resultValue.startsWith('ReferenceError: ') || resultValue.startsWith('TypeError: ')) && args.context !== 'repl') {
+            if (resultValue === "" || resultValue && (resultValue.startsWith('ReferenceError: ') || resultValue.startsWith('TypeError: ')) && args.context !== 'repl' ) {
```
digeff: Will this work if a variable is defined and it's value is undefined or null or false or an empty string?

PR #358, opened 2018-09-24
catching when watch variables don't exist

Shenniey:
This pull request is for the following issue that was posted to vscode-edge-debug2: https://github.com/Microsoft/vscode-edge-debug2/issues/68

If watch variable is not found, it will return error with message string instead of empty string.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -2729,7 +2729,10 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             } else {
                 propCountP = this.getCollectionNumPropsByEval(object.objectId);
             }
-        } else {
+        } else if (context === 'watch' && object.subtype && object.subtype === 'error') {
+            propCountP = Promise.reject(new Error(errors.evalNotAvailableMsg));
+        }
+        else {
             propCountP = Promise.resolve({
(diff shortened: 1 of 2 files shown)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -2729,7 +2729,10 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             } else {
                 propCountP = this.getCollectionNumPropsByEval(object.objectId);
             }
-        } else {
+        } else if (context === 'watch' && object.subtype && object.subtype === 'error') {
+            propCountP = Promise.reject(new Error(errors.evalNotAvailableMsg));
```
roblourens: This doesn't seem like the right place for this fix. This error normally comes from https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/chrome/chromeDebugAdapter.ts#L2492. Why does that not work for Edge? I can't repro this bug on Chrome.

Shenniey: I'm sorry, this is my mistake. The reason non-existent watch variables in edge don't get caught is because edge does not have the property exceptionDetails when it comes back from the browser. It just returned with results as an empty string. I should have checked that it was edge specific, my bad. I'm making the fix on the edge debug adapter now.

PR #360, opened 2018-09-26
 Fix breakpoint logic for multiple authored files sourcemapped to a single js bundle

mrcrane:
When multiple authored files are compiled to a single js bundle, the set breakpoint logic is broken.

Current broken logic: When setting breakpoints we clear all existing breakpoints in the target url (in this case, the js bundle), and then we apply the incoming breakpoints for the authored file. This wipes out any breakpoints for all other authored files that are sourcemapped to the same js bundle.

Proposed fix: Track the authored path for all breakpoints if it is available and only remove breakpoints that exist for that authored file.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -1461,13 +1461,18 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
                     targetScriptUrl = args.source.path;
                 }
 
+                let authoredPath: string;
+                if (args.authoredPath) {
+                    authoredPath = utils.canonicalizeUrl(args.authoredPath);
+                }
+
                 if (targetScriptUrl) {
                     // DebugProtocol sends all current breakpoints for the script. Clear all breakpoints for the script then add all of them
                     const internalBPs = args.breakpoints.map(bp => new InternalSourceBreakpoint(bp));
                     const setBreakpointsPFailOnError = this._setBreakpointsRequestQ
-                        .then(() => this.clearAllBreakpoints(targetScriptUrl, args.authoredPath))
+                        .then(() => this.clearAllBreakpoints(targetScriptUrl, authoredPath))
                         .then(() => this.addBreakpoints(targetScriptUrl, internalBPs))
-                        .then(responses => ({ breakpoints: this.targetBreakpointResponsesToBreakpointSetResults(targetScriptUrl, args.authoredPath, responses, internalBPs, ids) }));
+                        .then(responses => ({ breakpoints: this.targetBreakpointResponsesToBreakpointSetResults(targetScriptUrl, authoredPath, responses, internalBPs, ids) }));
 
                     const setBreakpointsPTimeout = utils.promiseTimeout(setBreakpointsPFailOnError, ChromeDebugAdapter.SET_BREAKPOINTS_TIMEOUT, localize('setBPTimedOut', 'Set breakpoints request timed out'));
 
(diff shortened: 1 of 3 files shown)


roblourens:
Thanks! I will take a look today.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -87,6 +87,11 @@ export interface IOnPausedResult {
     didPause: boolean;
 }
 
+export interface CommittedBreakpoint {
```
roblourens: `ICommittedBreakpoint`

test/chrome/chromeDebugAdapter.test.ts
```diff
@@ -472,9 +475,73 @@ suite('ChromeDebugAdapter', () => {
...
+            const breakpoints2: DebugProtocol.SourceBreakpoint[] = [
+                { line: 5, column: 6 },
+                { line: 7, column: 8 }
+            ];
+
+            const authoredSourcePath1 = '/project/authored1.ts';
+            const authoredSourcePath2 = '/project/authored2.ts';
+            const generatedScriptPath = '/project/bundle.js';
+
+            await chromeDebugAdapter.attach(ATTACH_ARGS)
```
roblourens: You can just use await for all of these

roblourens: And the below test

mrcrane:
I will create an issue with steps soon.

In the meantime, the unit tests that I wrote fail without this fix. Let me know if those tests aren't clear. I'm certainly open to the possibility that this does not go far enough, or that it is implemented in the wrong place.

The crux of the issue that I'm aware of is here:

https://github.com/Microsoft/vscode-chrome-debug-core/blob/674f0cb1098db36c8e1887cfb8f537dcd84ff376/src/chrome/chromeDebugAdapter.ts#L1448

If the sourceMapTransformer finds all BPs from other .ts files that map to the same .js file, I'm not exactly sure how best to use that information to clear and set breakpoints on the target.

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1205,6 +1220,11 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         };
         const scriptPath = this._pathTransformer.breakpointResolved(bp, script.url);
 
+        const authoredPosition = await this._sourceMapTransformer.mapToAuthored(scriptPath, bp.line, bp.column);
+        const authoredPath = authoredPosition ? authoredPosition.source : null;
```
digeff: Do we need to cannonicalize this path?

roblourens:
Is it possible that the tests are just expecting a certain sequence of events that your PR implements? Because the basic scenario definitely works for me, I have an app where I can set breakpoints in different .ts files that are merged into one .js file. I do vaguely remember an issue where we discussed getting rid of this the sourceMapTransformer logic I mentioned above, but I can't find that issue or remember the details of it, and it is very possible that there is some issue with this scenario.

Maybe just a race condition when we miss a BP in the time between when it's removed and when it's reset. I just want to be clear about what scenario we are trying to fix here.

mrcrane:
I tried putting together more minimal steps to reproduce this issue and I found this scenario working properly. So you are correct. Stepping through this I now understand how this is supposed to be working.

The problem that I was running into is indeed in the code that you pointed out in baseSourceMapTransformer. Looks like there is actually an issue with looking up breakpoints in other files due to a case mismatch in file paths. I will investigate further, but yes this should be a much simpler fix.

PR #362, opened 2018-09-29
Fix having 2 this on the locals scope

digeff:
Fix having 2 this on the locals scope.
For some scenarios we were seeing 2 "this" variables in the locals scope. This eliminates the second copy...

### DIFF
--- a/src/chrome/variables.ts
+++ b/src/chrome/variables.ts
@@ -63,8 +63,8 @@ export class ScopeContainer extends BaseVariableContainer {
     public expand(adapter: ChromeDebugAdapter, filter?: string, start?: number, count?: number): Promise<DebugProtocol.Variable[]> {
         // No filtering in scopes right now
         return super.expand(adapter, 'all', start, count).then(variables => {
-            if (this._thisObj) {
-                // If this is a scope that should have the 'this', prop, insert it at the top of the list
+            if (this._thisObj && !variables.find(v => v.name === 'this')) {
+                // If this is a scope that should have the 'this', prop, insert it at the top of the list unless it's already there
                 return this.insertRemoteObject(adapter, variables, 'this', this._thisObj);
             }
 


roblourens:
I don't see this issue, how can I repro it? We need to fix the real issue not just hide a variable named 'this'?

digeff:
I'm hitting this issue when creating a new Angular project with ASP.NET Core 2.1 and setting and hitting a breakpoint on ClientApp\src\app\counter\counter.component.ts, line 11 (You need to go to the counter page and click the button to hit the BP).

Originally I was seeing this:
image

After the fix I see this:
image

digeff:
Interesting, Chrome DevTools also shows two "this" variables:
image

roblourens:
What's the difference between them do they point to the same object id? Do they have the same props? Can you reduce that problem code at all?

digeff:
Filed chromium issue to understand what is happening here: https://bugs.chromium.org/p/chromium/issues/detail?id=890920

roblourens:
Thanks for filing that, let's see what they say. Can you put the test repo on github so they can get try it more easily? I have no idea how to "Create an Angular application in ASP.NET Core 2.1", they might not either. It's just JS, maybe you can even reduce it for them.

digeff:
More detailed instructions as to how to reproduce this issue in VS:
1.	Make sure that the ASP.NET and web development workload is installed in the Visual Studio Installer:
image

 2.	Open the new project wizard in VS:
image
 
3.	Select an ASP.NET Core Web Application and click OK:
image
 
4.	Select: .NET Core, ASP.NET Core 2.1, Angular, and unselect Configure for HTTPS, and then click OK:
image
 
5.	Set a breakpoint on ClientApp\src\app\counter\counter.component.ts on line 11.
image
 
6.	Start the debugger by pressing F5
7.	Go to the Counter tab and click the Increment button
 
image

8. Check the local variables when the breakpoint hit to see the issue:
image

phistuck:
It would be great if you added (here and in the Chrome issue tracker) a link to a small GitHub project (or a GitHub Gist) that has the output folder of your project (I expect an HTML file, a JavaScript bundle file and maybe a CSS bundle file and irrelevant images).
It will shorten the time it takes to reproduce and fix this, because the one who works on your issue will not be required to find a Windows computer, or install Visual Studio (plus, it will eliminate any server-related issues).

PR #364, opened 2018-10-01
Fix for file path case mismatch in sourcemaps

mrcrane:
New and improved one-line fix, as an alternative to #360 

When the sourcemaps are processed the paths are canonicalized with utils.canonicalizeUrl(). The path being passed in with setBreakpoints needs to be canonicalized as well in order for the paths to match properly.

It turns out this is actually only an issue with Visual Studio, since the paths are changed to lowercase only when the client is detected to be Visual Studio.

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -1405,7 +1405,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         this.reportBpTelemetry(args);
         if (args.source.path) {
             args.source.path = this.displayPathToRealPath(args.source.path);
-            args.source.path = utils.fixDriveLetterAndSlashes(args.source.path);
+            args.source.path = utils.canonicalizeUrl(args.source.path);
         }
 
         return this.validateBreakpointsPath(args)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -1405,7 +1405,7 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
         this.reportBpTelemetry(args);
         if (args.source.path) {
             args.source.path = this.displayPathToRealPath(args.source.path);
-            args.source.path = utils.fixDriveLetterAndSlashes(args.source.path);
```
mrcrane: I am removing fixDriveLetterAndSlashes() since that is called internally by canonicalizeUrl()

roblourens:
Does this mean that you can never set a breakpoint in a file with a capital letter in its path? I still don't understand exactly what is the issue that we are trying to fix here because these sounds like a completely different issue than what you were describing before.

mrcrane:
The issue is the same as before, although now I know that it will only happen with Visual Studio.

This line of code is supposed to find BPs from other authored files that map to the same file: https://github.com/Microsoft/vscode-chrome-debug-core/blob/3da45fba99d26705701316ce7f63c3f5678fb562/src/transformers/baseSourceMapTransformer.ts#L142

This is failing because of a case mismatch in the file paths.

The _sourceMaps paths are canonicalized here:
https://github.com/Microsoft/vscode-chrome-debug-core/blob/3da45fba99d26705701316ce7f63c3f5678fb562/src/sourceMaps/sourceMap.ts#L122

Since the file paths coming in from setBreakpoints args are not canonicalized, this path lookup will fail if the path contains upper case letters.

This is only affecting Visual Studio because paths are only converted to lowercase when the client is detected to be Visual Studio: https://github.com/Microsoft/vscode-chrome-debug-core/blob/3da45fba99d26705701316ce7f63c3f5678fb562/src/chrome/chromeDebugAdapter.ts#L242

PR #366, opened 2018-10-09
bumping version number to take in newer version of debug adapter

Shenniey:
Taking in changes to debug adapter that modify timestamp to UTC

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "vscode-chrome-debug-core",
   "displayName": "vscode-chrome-debug-core",
-  "version": "6.7.22",
+  "version": "6.7.21",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "vscode-chrome-debug-core",
   "displayName": "vscode-chrome-debug-core",
-  "version": "6.7.21",
+  "version": "6.7.22",
```
roblourens: Don't change this version, I'll do it when I publish it (which I can do right now but this is easier to manage if this version isn't changed in a PR)

Shenniey: Oh, ok! Sorry about that, undoing it now.

PR #367, opened 2018-10-12
Added support for basic color escape sequencies on Debug Console

rdegelo:
Added support for bold font style on Debug Console.

Background color can be easily implemented using the same approach.
If this PR gets accepted, I can add background colors support.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -107,6 +107,30 @@
         }
       }
     },
+    "@types/color": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/@types/color/-/color-3.0.0.tgz",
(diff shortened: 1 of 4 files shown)


roblourens:
This looks great, but the font-weight doesn't appear to have any affect in the Debug Console. Is it working for you? I don't know whether vscode supports it.

rdegelo:
It works for me: 
```
console.log('%c Normal font weight ', 'color: blue');
console.log('%c Normal font weight ', 'color: blue; font-weight: bold');
```

Result: 
image

To check what is supported, you can see this file: 
https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/parts/debug/browser/debugANSIHandling.ts#L70

roblourens:
I see it, the problem is that the debug console uses the Monaco font on macos, which doesn't support a bold font weight.

rdegelo:
Cool, 
When I have time I'll add background/underline support and post a new PR.

PR #368, opened 2018-10-15
Fix node not hitting breakpoints in network paths

digeff:
We couldn't hit a breakpoint in:
\\myshare\myfolder\myfile.js


roblourens:
> // Breakpoints need to be re-set when break-on-load is enabled

Why? We guessed at the path, that path was correct right? What's the real issue here? I don't see how breakOnLoad or network paths are involved.

digeff:
We guessed at the path, but we never set that breakpoint with that path (With break-on-load we set the break-on-load breakpoint instead of the real breakpoint, we need to set the real-breakpoint when parse the script).

PR #374, opened 2018-10-23
Improving Debug Console Styling. Adding underline and background support

rdegelo:


src/chrome/consoleHelper.ts
```diff
@@ -150,31 +147,41 @@ function formatArg(formatSpec: string, arg: Crdp.Runtime.RemoteObject): string |
     } else if (formatSpec === 'f') {
         return +paramValue + '';
     } else if (formatSpec === 'c') {
-        // value:"color: gray; font-weight: light
-        const colorMatches = colorRegex.exec(arg.value);
-        const fontWeightMatches = fontWeightRegex.exec(arg.value);
+        const cssRegex = /\s*(.*?)\s*:\s*(.*?)\s*(?:;|$)/g;
```
roblourens: Isn't the `.*?` redundant or is that different than `.*`?

rdegelo: Its not, ```*``` is greedy by default, and will try to match as much text as possible and will eventually match multiple parameters. ```?``` makes it lazy.

```
test: aaaa; test2: bbbbb;
```

If ```?``` gets removed, the first match on this text will be the whole string, and that's not what we want.

roblourens: Oh, didn't know that, thanks.

PR #379, opened 2018-11-12
cannonicalizing url so that it can find the matching script when resolving breakpoint

Shenniey:
the breakpoint is moving b/c it cannot match authored to generated path in node scenario: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/685518

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -1203,23 +1203,13 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             return;
         }
 
-        // If we cannot find the script url in the committed breakpoints, we check for the canonicalized path. If found, use those committed breakpoints
-        // If we cannot find either the given url or the canonicalized url (or they are the same and the script simply cannot be found), committed breakpoints is an empty array
-        let committedBps = this._committedBreakpointsByUrl.get(script.url);
-        if (!committedBps) {
-            let canonicalizedUrl = utils.canonicalizeUrl(script.url);
-            committedBps = this._committedBreakpointsByUrl.get(canonicalizedUrl);
-            if (committedBps) {
-                script.url = canonicalizedUrl;
-            } else {
-                committedBps = [];
-            }
-        }
+        // committed breakpoints (this._committedBreakpointsByUrl) should always have url keys in canonicalized form
+        let committedBps = this.getValueFromCommittedBreakpointsByUrl(script.url);
 
         if (!committedBps.find(committedBp => committedBp.breakpointId === params.breakpointId)) {
             committedBps.push({breakpointId: params.breakpointId, actualLocation: params.location});
         }
-        this._committedBreakpointsByUrl.set(script.url, committedBps);
+        this.setValueForCommittedBreakpointsByUrl(script.url, committedBps);
 
         const bp = <DebugProtocol.Breakpoint>{
             id: breakpointId,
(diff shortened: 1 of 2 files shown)


roblourens:
I can see the upstream issue but that issue tracker isn't public, could you copy it into this repo?

Also, if you debugged this could you show why the url needs to be canonicalized here? Is this another file: URI case?

Shenniey:
Hi, I'm sorry for doing this so late, but yes, I've added the issue I to this repo: https://github.com/Microsoft/vscode-chrome-debug-core/issues/380

Yes, it's another file:URI case. The problem shows up in OnBreakpointsResolved, when it is looking for a matching breakpoint for breakpoints in routing files in commited breakpoints. The filepath in committed breakpoints is canonicalized (though the main app.js file is not canonicalized...)
committed breakpoints by url

But the url it is searching for has the file:/// prefix:

lookingfor script

I am going to make another commit because although in the instances I've seen so far, only breakpoints in routing files are searched for in OnBreakpointResolved (likely due to lazy loading, although I'm not 100% sure), so the fact that the main app.js file is not canonicalized has not caused any issues in my own testing. However, I'm not sure if that would come up as a problem anywhere else, so I am planning to change it so that it checks in the committed breakpoints for the given file path, and then checks for the canonicalized filepath if it cannot find the given one. Would that be a better approach?

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1203,7 +1203,19 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
...
 
-        const committedBps = this._committedBreakpointsByUrl.get(script.url) || [];
+        // If we cannot find the script url in the committed breakpoints, we check for the canonicalized path. If found, use those committed breakpoints
+        // If we cannot find either the given url or the canonicalized url (or they are the same and the script simply cannot be found), committed breakpoints is an empty array
+        let committedBps = this._committedBreakpointsByUrl.get(script.url);
+        if (!committedBps) {
+            let canonicalizedUrl = utils.canonicalizeUrl(script.url);
+            committedBps = this._committedBreakpointsByUrl.get(canonicalizedUrl);
+            if (committedBps) {
+                script.url = canonicalizedUrl;
```
roblourens: I'm not sure about mutating something that was already saved in scriptsById. If it should be canonicalized at this point, then that should be done at the point it was saved into scriptsById

roblourens:
Thanks for the details!

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1203,7 +1203,19 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             return;
         }
 
-        const committedBps = this._committedBreakpointsByUrl.get(script.url) || [];
+        // If we cannot find the script url in the committed breakpoints, we check for the canonicalized path. If found, use those committed breakpoints
+        // If we cannot find either the given url or the canonicalized url (or they are the same and the script simply cannot be found), committed breakpoints is an empty array
+        let committedBps = this._committedBreakpointsByUrl.get(script.url);
```
digeff: If the scripts inside are canonicalized we should be able to just search for the canonicalized version

src/chrome/chromeDebugAdapter.ts
```diff
@@ -1203,19 +1203,23 @@ export abstract class ChromeDebugAdapter implements IDebugAdapter {
             return;
         }
 
-        const committedBps = this._committedBreakpointsByUrl.get(script.url) || [];
+        // committed breakpoints (this._committedBreakpointsByUrl) should always have url keys in canonicalized form
+        const committedBps = this.getValueFromCommittedBreakpointsByUrl(script.url);
```
digeff: What will happen if there is no commitedBps for this url?

Shenniey: Sorry I forgot to put back the empty array

PR #381, opened 2018-11-19
making sure bp.actual location is not null/undefined before trying to access its line number

Shenniey:
Chrome core has a known issue where occasionally it returns an empty locations array: https://github.com/ChromeDevTools/devtools-protocol/issues/103

When this happens, actual location is undefined and throws an error, causing the browser to hang without loading the page. Stack trace from logs:
"aggregated.exceptionStack":"[
  \"TypeError: Cannot read property 'lineNumber' of undefined\\n    
  at committedBps.filter.bp (breakOnLoadHelper.js:100:96)\\n    
  at Array.filter (<anonymous>)\\n    
  at BreakOnLoadHelper.<anonymous> (breakOnLoadHelper.js:100:65)\\n    
  at Generator.next (<anonymous>)\\n    at fulfilled (breakOnLoadHelper.js:7:58)\\n    
  at process._tickCallback (internal/process/next_tick.js:68:7)\"
]"

This change is just to make sure that the browser doesn't hang

### DIFF
--- a/src/chrome/breakOnLoadHelper.ts
+++ b/src/chrome/breakOnLoadHelper.ts
@@ -113,7 +113,10 @@ export class BreakOnLoadHelper {
         // Important: We need to get the committed breakpoints only after all the pending breakpoints for this file have been resolved. If not this logic won't work
         const committedBps = this._chromeDebugAdapter.committedBreakpointsByUrl.get(pausedScriptUrl) || [];
         const anyBreakpointsAtPausedLocation = committedBps.filter(bp =>
-            bp.actualLocation.lineNumber === pausedLocation.lineNumber && bp.actualLocation.columnNumber === pausedLocation.columnNumber).length > 0;
+            // Important: Chrome devtools-protocol can return an empty locations array, resulting in 'undefined' for the actualLocation, so we have to check an actualLocation exists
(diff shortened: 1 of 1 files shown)


roblourens:
@digeff can you review?

roblourens:
The locations array will be empty any time the breakpoint isn't bound to a location, so is this really a bug on their end, or just another case for us to handle?

Shenniey:
I am leaning more towards bug, but I'm not certain. From the issue it seemed as though the locations array shouldn't be empty. Also, when I was testing with some of the breakpoint locations where we are getting back empty location arrays, if I run the project without debugging through the editor and just use F12 in Chrome, the same breakpoints successfully bind to a location and pause at those locations.

Also, if I debug with Edge, we don't have this issue.

digeff:
I'm slightly surprised that not all the elements of commitedBps have a valid location. I think this is a valid way to avoid the crash.

The other option we could consider is to modify: https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/chrome/chromeDebugAdapter.ts#L1667 so that we don't consider the breakpoints commited if it doesn't have a valid location. I'm not sure what the consequences of that change would be, so I think that the current approach is okay.

PR #384, opened 2018-12-03
Core files of the script model

digeff:
These are the core files of the script model that we are creating/refactoring for the V2 version of the debug adapters.

### DIFF
--- a/src/chrome/internal/scripts/sourcesMapper.ts
+++ b/src/chrome/internal/scripts/sourcesMapper.ts
@@ -3,8 +3,8 @@ import { SourceMap } from '../../../sourceMaps/sourceMap';
 
 export interface ISourcesMapper {
     readonly sources: string[];
-    getSourcePosition(positionInScript: IPositionInScript): IPositionInSource | null;
-    getScriptPosition(positionInSource: IPositionInSource): IPositionInScript | null;
+    getPositionInSource(positionInScript: IPositionInScript): IPositionInSource | null;
+    getPositionInScript(positionInSource: IPositionInSource): IPositionInScript | null;
 }
 
 interface IPositionInSource {
@@ -20,14 +20,14 @@ interface IPositionInScript {
 
 /** This class maps locations from a script into the sources form which it was compiled, and back. */
 export class SourcesMapper implements ISourcesMapper {
-    public getSourcePosition(positionInScript: IPositionInScript): IPositionInSource | null {
+    public getPositionInSource(positionInScript: IPositionInScript): IPositionInSource | null {
         const mappedPosition = this._sourceMap.authoredPositionFor(positionInScript.line, positionInScript.column || 0);
         return mappedPosition && mappedPosition.source && mappedPosition.line
             ? { source: mappedPosition.source, line: mappedPosition.line as LineNumber, column: mappedPosition.column as ColumnNumber }
             : null;
     }
 
-    public getScriptPosition(positionInSource: IPositionInSource): IPositionInScript | null {
+    public getPositionInScript(positionInSource: IPositionInSource): IPositionInScript | null {
         const mappedPosition = this._sourceMap.generatedPositionFor(positionInSource.source,
             positionInSource.line, positionInSource.column || 0);
         return mappedPosition && mappedPosition.line
@@ -44,11 +44,11 @@ export class SourcesMapper implements ISourcesMapper {
 }
 
 export class NoSourceMapping implements ISourcesMapper {
-    public getSourcePosition(_: IPositionInScript): null {
+    public getPositionInSource(_: IPositionInScript): null {
         return null;
     }
 
-    public getScriptPosition(_: IPositionInSource): null {
+    public getPositionInScript(_: IPositionInSource): null {
         return null;
     }
 


roblourens:
Please include some more context with these PRs. Just a couple sentences about what is going on here so it isn't a total mystery to someone who isn't you or I.

digeff:
My hope is that the descriptions that I added to the main interfaces and classes will give that context, so anyone who can see the code in the future can easily find that information too.

roblourens:
Sure, I just mean, there should be a basic explanation of "what is the point of this PR" that someone can see without reading code.

auchenberg:
@digeff Good post on PRs, https://blog.codeminer42.com/on-writing-a-great-pull-request-37c60ce6f31d

roblourens:
I agree that you shouldn't be writing an essay or documenting your code in a PR description. I'm just asking that the PR have some explanation of, what is the point of this PR? What you've edited it to is already an improvement. Someone just looking at this PR for the first time should know whats in it without having to read code.

digeff:
I agree that my first description was too brief. Is there any particular information that you think is missing now?

src/chrome/internal/scripts/script.ts
```diff
@@ -0,0 +1,117 @@
...
+import { IExecutionContext } from './executionContext';
+
+/** This interface represents a piece of code that is being executed in the debugee. Usually a script matches to a file or a url, but that is not always the case.
+ * This interface solves the problem of finding the different loaded sources associated with a script, and being able to identify and compare both scripts and sources easily.
+ */
+export interface IScript {
+    readonly executionContext: IExecutionContext;
+    readonly runtimeSource: ILoadedSource<CDTPScriptUrl>; // Source in Webserver
+    readonly developmentSource: ILoadedSource; // Source in Workspace
+    readonly sourcesOfCompiled: SourceOfCompiled[]; // Sources before compilation
```
mrcrane: What's the difference between a development source and a "source of compiled"? Aren't the development sources the ones that are compiled?

src/chrome/internal/scripts/sourcesMapper.ts
```diff
@@ -0,0 +1,58 @@
+import { LineNumber, ColumnNumber } from '../locations/subtypes';
+import { SourceMap } from '../../../sourceMaps/sourceMap';
+
+export interface ISourcesMapper {
+    readonly sources: string[];
+    getSourceLocation(scriptLocation: IPositionInScript): IPositionInSource | null;
+    getScriptLocation(scriptLocation: IPositionInSource): IPositionInScript | null;
```
mrcrane: The terms SourceLocation and ScriptLocation seem a bit ambiguous. You have references to "runtimeSource" in script.ts for example.

Also the source mapper already has a convention of calling them "authored" sources and "generated" sources which make sense to me. Any reason we don't stick to those terms, here and in the rest of the code?

src/chrome/internal/sources/loadedSource.ts
```diff
@@ -0,0 +1,101 @@
...
+import { IResourceIdentifier, parseResourceIdentifier, ResourceName } from './resourceIdentifier';
+import { determineOrderingOfStrings } from '../../collections/utilities';
+
+/** This interface represents a source or text that is related to a script that the debugee is executing. The text can be the contents of the script itself,
+ *  or a file from which the script was loaded, or a file that was compiled to generate the contents of the script
+ */
+export interface ILoadedSource<TString = string> {
+    readonly script: IScript;
+    readonly identifier: IResourceIdentifier<TString>;
+    readonly origin: string;
```
mrcrane: What does origin refer to? Strange to see just a bare string type in here 😄

src/chrome/internal/sources/resourceIdentifier.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+/**
+ * Sample formats:
+ * file:///D:\\scripts\\code.js
+ * file:///Users/me/project/code.js
+ * c:/scripts/code.js
+ * http://site.com/scripts/code.js
+ * http://site.com/
```
mrcrane: Would be great to see unit tests covering these cases.

src/chrome/internal/scripts/script.ts
digeff: A developmentSource is the .js location inside the user workspace
source of compiled now renamed to mapped source is the sources that we get out of the SourceMap such as typescript sources.

src/chrome/internal/scripts/sourcesMapper.ts
digeff: SourceLocation usually refers to the data that we get from the client, which might need or not to be source mapped. ScriptLocation refers to the locations relative to the script in CDTP so no mapping is needed.

The reason I don't use authored is because the way I understand it authored refers to .ts when you have them, and to the traditional .js when you don't. This concept was meant to refer to .ts only, and empty if we don't have any.

src/chrome/internal/sources/loadedSource.ts
digeff: This property serves the purpose of the origin property in the VS Code protocol: https://github.com/Microsoft/vscode-debugadapter-node/blob/64665fe18a0316b2a871df4b36cff214c28d2598/protocol/src/debugProtocol.ts#L1315

We'll use it to explain where we got the sources from (memory, source map, disk, etc...)

src/chrome/internal/sources/resourceIdentifier.ts
digeff: We are definitely going to add unit tests for these cases. We are planning on adding them after V2 stabilizes, and we don't have any remaining big changes to do to it.

roblourens:
My only suggestion right now is that we not rely on strings for paths, but that the fundamental type be URIs instead. That will help with

- Breakpoints in non-local files set from the client
- Not having to convert between file URI and path
- Not worrying about slash direction

digeff:
I think it's a great idea to use that library to handle all URIs related logic (URL class and subclasses).
I'm not convinced of the advantages of using it to handle the file paths related logic (LocalFilePath and subclasses)

roblourens:
Sure. Should I merge this now?

src/chrome/internal/scripts/sourcesMapper.ts
mrcrane: Actually there's a minor typo here,

`getScriptLocation(scriptLocation: IPositionInSource)`

variable name is wrong

mrcrane: ...and actually the methods and parameters should be named `getSourcePosition`, `getScriptPosition`, etc. to be consistent with the types and to reduce confusion with the url of the source which is also considered a location

digeff: Changed

digeff:
Yes, please merge it :)

PR #387, opened 2019-01-02
Created the UnresolvedSource to handle the source parameters sent via the vs code protocol that can be a path or a sourceReference

digeff:
This is part of the V2 refactoring of the debug adapter.

Created the UnresolvedSource to handle the source parameters sent via the vs code protocol that can be a path or a sourceReference.

When we need to use the parameter we'll call unresolvedSource.tryResolving which uses continuation passing style ( https://en.wikipedia.org/wiki/Continuation-passing_style ) to see if the source can be resolved or not, and take the appropriated action.

### DIFF
--- a/src/chrome/internal/sources/sourceResolver.ts
+++ b/src/chrome/internal/sources/sourceResolver.ts
@@ -21,7 +21,7 @@ export class SourceResolver implements IComponent {
 
     public tryResolving<R>(sourceIdentifier: IResourceIdentifier,
         succesfulAction: (resolvedSource: ILoadedSource) => R,
-        failedAction: (sourceIdentifier: IResourceIdentifier) => R = path => { throw new Error(`Couldn't find the source at path ${path}`); }): R {
+        failedAction: (sourceIdentifier: IResourceIdentifier) => R): R {
         const source = this._pathToSource.tryGetting(sourceIdentifier);
         if (source !== undefined) {
             return succesfulAction(source);
@@ -30,7 +30,7 @@ export class SourceResolver implements IComponent {
         }
     }
 
-    public createSourceResolver(sourceIdentifier: IResourceIdentifier): IUnresolvedSource {
+    public createUnresolvedSource(sourceIdentifier: IResourceIdentifier): IUnresolvedSource {
         return new SourceToBeResolvedViaPath(sourceIdentifier, this);
     }
 
--- a/src/chrome/internal/sources/sourceResolver.ts
+++ b/src/chrome/internal/sources/sourceResolver.ts
@@ -1,5 +1,5 @@
 import { ILoadedSource } from './loadedSource';
-import { IUnresolvedSource, SourceToBeResolvedViaPath } from './unresolvedSource';
+import { ISource, SourceToBeResolvedViaPath } from './source';
 import { newResourceIdentifierMap, IResourceIdentifier } from './resourceIdentifier';
 import { IComponent } from '../features/feature';
 import { ScriptParsedEvent } from '../../target/events';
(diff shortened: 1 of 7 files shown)


src/chrome/internal/sources/sourceResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+        failedAction: (sourceIdentifier: IResourceIdentifier) => R = path => { throw new Error(`Couldn't find the source at path ${path}`); }): R {
+        const source = this._pathToSource.tryGetting(sourceIdentifier);
+        if (source !== undefined) {
+            return succesfulAction(source);
+        } else {
+            return failedAction(sourceIdentifier);
+        }
+    }
+
+    public createSourceResolver(sourceIdentifier: IResourceIdentifier): IUnresolvedSource {
```
roblourens: creating an unresolved source, not a source resolver

src/chrome/internal/sources/sourceResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * The SourceResolver listens to onScriptParsed events to build a map of paths to loaded sources. When an SourceToBeResolvedViaPath is created, it'll store a reference to this object,
+ * and use it when it tries to resolve the path to a loaded source
+ */
+
+ @injectable()
+export class SourceResolver implements IComponent {
+    private _pathToSource = newResourceIdentifierMap<ILoadedSource>();
+
+    public tryResolving<R>(sourceIdentifier: IResourceIdentifier,
```
roblourens: Why use this style with the callbacks, vs just throw in case of failure?

src/chrome/internal/sources/unresolvedSource.ts
```diff
@@ -0,0 +1,56 @@
...
+        return `Resolve source via #${this.sourceIdentifier}`;
+    }
+
+    constructor(public readonly sourceIdentifier: IResourceIdentifier, private readonly _sourceResolver: SourceResolver) {
+        super();
+    }
+}
+
+// This source was already loaded, so we store it in this class
+export class SourceAlreadyResolvedToLoadedSource extends UnresolvedSourceCommonLogic implements IUnresolvedSource {
```
roblourens: I don't understand why there is a "SourceAlreadyResolved" which is an "IUnresolvedSource", where would this be used?

src/chrome/internal/sources/sourceResolver.ts
digeff: I expect most callers of IUnresolvedSource.tryResolving which is what's calling SourceResolver.tryResolving to have to handle the failure case immediately instead of letting it bubble-up to the top level. By using this signature for the method, I'm forcing the consumers to have to supply that failure case (So they can't forget to do it). If I expected most consumers to want to let the failure bubble-up so the top-level can handle it, I'd throw an Exception instead.

BTW: No-one is using the default parameter of failedAction so I'm removing it...

src/chrome/internal/sources/unresolvedSource.ts
digeff: You are right, it doesn't make sense.

I'd love your help coming up with some better names.

We use SourceAlreadyResolvedToLoadedSource when the client sends a sourceReference parameter. We use SourceToBeResolvedViaPath when the client sends a path parameter.

Maybe I should rename IUnresolvedSource to ISourceThatMayBeUnresolved?

Suggestions are welcome....

src/chrome/internal/sources/sourceResolver.ts
roblourens: It seems like the caller should be the one to decide whether the failure case is handled or not. But not a big deal.

src/chrome/internal/sources/unresolvedSource.ts
roblourens: If it's sometimes unresolved and sometimes resolved then it seems like it's just a "Source".

digeff: Renamed to ISource...

roblourens:
Ok to merge?

PR #388, opened 2019-01-08
Location model

digeff:
New location model for V2.
We use the location model to represent the location in a file/resources, e.g.: a breakpoint in myFile.js:34. We also use it to map those locations to other files using source maps eg.: myFile.js:34 --> myWidget.ts:26
We primarily use this location model to store and map information about breakpoints and stack traces between JavaScript and TypeScript.

### DIFF
--- a/src/chrome/internal/locations/location.ts
+++ b/src/chrome/internal/locations/location.ts
@@ -1,0 +1,157 @@
+import * as Validation from '../../../validation';
+import { IScript } from '../scripts/script';
+import { ISource } from '../sources/source';
+import { ILoadedSource } from '../sources/loadedSource';
+import { logger } from 'vscode-debugadapter';
+import { ColumnNumber, LineNumber, URLRegexp } from './subtypes';
+import { CDTPScriptUrl } from '../sources/resourceIdentifierSubtypes';
+import { IResourceIdentifier, parseResourceIdentifier, URL } from '../sources/resourceIdentifier';
+
+export type integer = number;
+
+export class Position {
+    constructor(
+        public readonly lineNumber: LineNumber,
+        public readonly columnNumber?: ColumnNumber) {
+        Validation.zeroOrPositive('Line number', lineNumber);
+        if (columnNumber !== undefined) {
+            Validation.zeroOrPositive('Column number', columnNumber);
+        }
+    }
+
+    public isSameAs(location: Position): boolean {
+        return this.lineNumber === location.lineNumber
+            && this.columnNumber === location.columnNumber;
+    }
+
+    public toString(): string {
+        return this.columnNumber !== undefined
+            ? `${this.lineNumber}:${this.columnNumber}`
+            : `${this.lineNumber}`;
+    }
+}
+
+interface ILocation<T extends ScriptOrSourceOrURLOrURLRegexp> {
+    readonly lineNumber: integer;
+    readonly columnNumber?: integer;
+    readonly coordinates: Position;
+    readonly resource: T;
+}
+
+export type ScriptOrSourceOrURLOrURLRegexp = IScript | ILoadedSource | ISource | URLRegexp | URL<CDTPScriptUrl>;
+
+export type Location<T extends ScriptOrSourceOrURLOrURLRegexp> =
+    T extends ISource ? LocationInSource : // Used when receiving locations from the client
+    T extends ILoadedSource ? LocationInLoadedSource : // Used to translate between locations on the client and the debugee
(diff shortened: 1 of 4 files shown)


src/chrome/internal/locations/location.ts
```diff
@@ -0,0 +1,157 @@
...
+            && this.columnNumber === location.columnNumber;
+    }
+
+    public toString(): string {
+        return this.columnNumber !== undefined
+            ? `${this.lineNumber}:${this.columnNumber}`
+            : `${this.lineNumber}`;
+    }
+
+    constructor(
```
roblourens: Minor point of style, I prefer putting the constructor at the top. Especially helpful so I can see what the members are first before looking at methods.

src/chrome/internal/locations/location.ts
```diff
@@ -0,0 +1,157 @@
...
+        public readonly lineNumber: LineNumber,
+        public readonly columnNumber?: ColumnNumber) {
+        Validation.zeroOrPositive('Line number', lineNumber);
+        if (columnNumber !== undefined) {
+            Validation.zeroOrPositive('Column number', columnNumber);
+        }
+    }
+}
+
+interface ILocation<T extends ScriptOrSourceOrURLOrURLRegexp> {
```
roblourens: What is this? It has coordinates and another line/col?

src/chrome/internal/locations/location.ts
```diff
@@ -0,0 +1,157 @@
...
+import { ISource } from '../sources/source';
+import { ILoadedSource } from '../sources/loadedSource';
+import { logger } from 'vscode-debugadapter';
+import { ColumnNumber, LineNumber, URLRegexp } from './subtypes';
+import { CDTPScriptUrl } from '../sources/resourceIdentifierSubtypes';
+import { IResourceIdentifier, parseResourceIdentifier, URL } from '../sources/resourceIdentifier';
+
+export type integer = number;
+
+export class Coordinates {
```
roblourens: Could we match vscode on this and call it a `Position`?

roblourens:
Make sure you fix your git email address so your commits are probably attributed to your account

src/chrome/internal/locations/location.ts
```diff
@@ -0,0 +1,157 @@
...
+    public toString(): string {
+        return `${this.resource}:${this.coordinates}`;
+    }
+
+    constructor(
+        public readonly resource: T,
+        public readonly coordinates: Coordinates) { }
+}
+
+export class LocationInSource extends LocationCommonLogic<ISource> implements ILocation<ISource> {
```
roblourens: `implements` should be unnecessary on all of these because LocationCommonLogic implements it

src/chrome/internal/locations/location.ts
digeff: @roblourens Do you want to add a lint rule that validates that the constructors follow that convention?

src/chrome/internal/locations/location.ts
digeff: The way I'm using these they are just an alias of the coordinates...
If you feel strongly about it, I can remove them and leave the coordinates only...

src/chrome/internal/locations/location.ts
digeff: Sure. I'll rename it.

src/chrome/internal/locations/location.ts
roblourens: Sure but I don't feel that strongly about it.

digeff: I'll try my best to remember it then...

src/chrome/internal/locations/location.ts
roblourens: Oh, ok. I think that's not really obvious from the interface. I think the interface should either expose the line/col or the coordinates, but not both. I don't really care which way.

src/chrome/internal/locations/location.ts
digeff: Removed

src/chrome/internal/locations/location.ts
digeff: Just to confirm: Do you also want to remove them from the classes? Or just the interface? (I'd like to leave them on the classes)

roblourens: Leaving them on the class as a private thing is ok. But the interface should make it clear how it's supposed to be used.

digeff: I'll remove them from the classes then. Thanks!

PR #391, opened 2019-01-15
Stack traces model

digeff:
The model classes represent the information about stack traces. The CDTP layer returns all the information in Scripts and the feature layer maps them to LoadedSources.
CDTP also has 2 different types of StackTraces. Runtime StackTraces, and Debugger stack traces. Runtime ones only have code flow information. Debugger ones also have state information. We represent that by CodeFlowFrame (It shows which code is being executed) having the shared code flow information, and CallFrame (It has the state associated with the function call) having the state information that only the Debugger stacks have.

The presentation classes are what we use when we want to show a stack trace to the user, and we convert it to an appropriated format to show it.

build.cmd: It's a utility I use to build and test my changes on Windows


build.cmd
```diff
@@ -0,0 +1,2 @@
+@gulp _dev-build
```
roblourens: I don't see why we should have a script that just invokes a single gulp task on windows, let's not check it in..

digeff: Are you okay with me adding build.cmd to .gitignore, so I can keep it on my folder?

roblourens: You can put it in `.git/info/exclude` which is like a private gitignore

digeff: I didn't know that was a thing... Thanks! :)

src/chrome/internal/stackTraces/stackTracePresentation.ts
```diff
@@ -0,0 +1,11 @@
+import { StackTracePresentationRow } from './stackTracePresentationRow';
+
+/** The stack traces we sent to the client will be represented by this classes and it is a combination of:
+ *    1. CallFrames with state information from the sync frames.
+ *    2. CodeFlowFrames without state information from async frames.
+ *    3. Labels that we use to [Show more frames] or [Frames skipped by smartStep], etc...
+ */
+ export interface StackTracePresentation {
```
roblourens: I'm confused about all the `Presentation` types. This and StackTracePresentationRow both say they are "sent to the client" but they don't really look like the DA type

src/chrome/internal/stackTraces/stackTracePresentation.ts
```diff
@@ -0,0 +1,11 @@
+import { StackTracePresentationRow } from './stackTracePresentationRow';
+
+/** The stack traces we sent to the client will be represented by this classes and it is a combination of:
+ *    1. CallFrames with state information from the sync frames.
+ *    2. CodeFlowFrames without state information from async frames.
+ *    3. Labels that we use to [Show more frames] or [Frames skipped by smartStep], etc...
```
roblourens: Neither of those are labels that come from the DA, only the async frame labels come from the DA

src/chrome/internal/stackTraces/codeFlowStackTrace.ts
```diff
@@ -0,0 +1,12 @@
+import { CodeFlowFrame } from './callFrame';
+import { IScript } from '../scripts/script';
+
+/** This class represents a stack trace that only has code flow information, but no state
```
roblourens: Another minor style point, I think the comment should start on the line after `/**`, e.g.

```
/**
 * comment
 */

File: src/chrome/internal/stackTraces/callFrame.ts [hunk 5]
```diff
@@ -0,0 +1,129 @@
...
+export type CallFrame<TResource extends ScriptOrLoadedSource> =
+    TResource extends ILoadedSource ? LoadedSourceCallFrame :
+    TResource extends IScript ? ScriptCallFrame :
+    ICallFrame<never>; // TODO: Figure out how to change this for never
+
+abstract class CallFrameCommonLogic<TResource extends ScriptOrLoadedSource> implements ICallFrame<TResource> {
+    public abstract get scopeChain(): Scope[];
+    public abstract get codeFlow(): CodeFlowFrame<TResource>;
+
+    public get source(): TResource extends ILoadedSource ? TResource : never {
```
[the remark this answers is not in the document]
digeff: Could you please give me more details about what makes this confusing?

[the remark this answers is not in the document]
roblourens: If I am looking at properties available on the object, I see that there is `source` and also that there is a `codeFlow` which has a `source` inside it, and it's not obvious what the difference is, or whether they are the same, and which one I should use.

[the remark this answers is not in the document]
digeff: If I added some comments that explained what each property is, and clearly say that both sources are the same, would that address your concern?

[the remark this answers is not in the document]
roblourens: Not really, why do we need two ways to access the same info?

src/chrome/internal/stackTraces/callFrame.ts
```diff
@@ -0,0 +1,129 @@
...
+    public get index(): number {
+        return this.codeFlow.index;
+    }
+
+    public get functionDescription(): string {
+        return this.codeFlow.functionDescription;
+    }
+}
+
+export class ScriptCallFrame extends CallFrameCommonLogic<IScript> {
```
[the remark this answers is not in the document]
roblourens: It's not just the naming, I'm not sure how they are used. `LoadedSourceCallFrame` contains a `ScriptCallFrame` and also a `CodeFlowFrame`, and I'm just not really sure what that "means". Like now there are 3 different views of the same frame in play.

src/chrome/internal/stackTraces/callFrame.ts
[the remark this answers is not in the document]
digeff: I think that both methods/properties make sense. I might need the source associated with a frame, or the code flow. When I need the source I'll naturally do frame.source rather than try to remember which instance variable has the source. It also avoids witting frame.codeFlow.source several times in the code.

Another good example (Although I'm going to remove this after the html scripts update) is script.url which is defined as script.runtimeSource.identifier.textRepresentation. If I want to get the url of a script it made sense in our older model to do script.url rather than have to remember to do: script.runtimeSource.identifier.textRepresentation.
Also when I read script.runtimeSource.identifier.textRepresentation it takes me longer to understand what is happening compared to just reading script.url.

I think a lot of times it's useful to have different ways to access the same information.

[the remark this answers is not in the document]
roblourens: Can the `codeFlow` be private?

roblourens:
I just remembered something else, we need copyright notices at the top of all files. But we can ignore that for now and add them all in one batch later, if that helps you stay flexible in refactoring.

src/chrome/internal/stackTraces/callFrame.ts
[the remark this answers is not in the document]
digeff: I've just checked that and we use it indirectly on the FramePresentationCommonLogic. Currently we access the properties like:

`
framePresentation.source and
framePresentation.lineNumber
`
And I also have some logic that makes it so that both CodeFlow and CallFrame use the same interface (which uses the public codeflow property).

If I remove all the "duplicated sources" we'll end up with some code like this:
`framePresentation.isCallFrame() ? framePresentation.callFrame.codeFlow.location.position.lineNumber :  framePresentation.codeFlow.location.position.lineNumber
`
instead of just writing: 
`framePresentation.lineNumber
`

So I'd definitely love to keep the duplicated info...

src/chrome/internal/stackTraces/callFrame.ts
```diff
@@ -0,0 +1,129 @@
...
+    readonly frameThis?: CDTP.Runtime.RemoteObject;
+    readonly returnValue?: CDTP.Runtime.RemoteObject;
+}
+
+export type CallFrame<TResource extends ScriptOrLoadedSource> =
+    TResource extends ILoadedSource ? LoadedSourceCallFrame :
+    TResource extends IScript ? ScriptCallFrame :
+    ICallFrame<never>; // TODO: Figure out how to change this for never
+
+abstract class CallFrameCommonLogic<TResource extends ScriptOrLoadedSource> implements ICallFrame<TResource> {
```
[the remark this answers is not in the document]
digeff: Changed

src/chrome/internal/stackTraces/callFrameDescription.ts
```diff
@@ -0,0 +1,38 @@
...
+
+            if (this._formatArgs.line) {
+                formattedDescription += ` Line ${locationInLoadedSource.position.lineNumber}`;
+            }
+        }
+
+        return formattedDescription;
+    }
+
+    constructor(private readonly _callFrame: ScriptCallFrame, private readonly _formatArgs?: DebugProtocol.StackFrameFormat) { }
```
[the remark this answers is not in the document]
digeff: Changed

src/chrome/internal/stackTraces/callFramePresentation.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    public get lineNumber(): number {
+        return this.codeFlow.lineNumber;
+    }
+
+    public get columnNumber(): number {
+        return this.codeFlow.columnNumber;
+    }
+
+    public isNotLabel(): this is ICodeFlowFramePresentation {
```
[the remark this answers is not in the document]
digeff: Changed to isCodeFlow

src/chrome/internal/stackTraces/stackTracePresentationRow.ts
```diff
@@ -0,0 +1,25 @@
+import { ICodeFlowFramePresentation, CallFramePresentation } from './callFramePresentation';
+
+export type CallFramePresentationHint = 'normal' | 'label' | 'subtle';
+
+// Row of a stack trace that we send to the client
+export interface StackTracePresentationRow {
```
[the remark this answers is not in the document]
digeff: Changed

src/chrome/internal/stackTraces/codeFlowStackTrace.ts
digeff: Changed

src/chrome/internal/stackTraces/callFrame.ts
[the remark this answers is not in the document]
roblourens: Ehhhhh ok

src/chrome/internal/stackTraces/callFramePresentation.ts
```diff
@@ -0,0 +1,94 @@
...
+    public isCallFrame(): this is CallFramePresentation {
+        return true;
+    }
+
+    public get description(): string {
+        return this._descriptionFormatter.description;
+    }
+}
+
+export class CodeFlowFramePresentation extends BaseFramePresentation implements IStackTracePresentationRow {
```
roblourens: `IStackTracePresentationRow` is already implemented by the super class

src/chrome/internal/stackTraces/callFramePresentation.ts
```diff
@@ -0,0 +1,94 @@
...
+        additionalPresentationDetails?: ICallFramePresentationDetails,
+        presentationHint?: CallFramePresentationHint) {
+        super(additionalPresentationDetails, presentationHint);
+    }
+
+    public get description(): string {
+        return this.codeFlow.functionDescription;
+    }
+
+    public isCallFrame(): this is CallFramePresentation {
```
roblourens: It seems weird to have a method to test whether something is an instance of some other class, just use `instanceof`?

src/chrome/internal/stackTraces/stackTracePresentation.ts
roblourens: Ok. `I` name?

src/chrome/internal/stackTraces/formatCallFrameDescription.ts
```diff
@@ -0,0 +1,25 @@
+import * as path from 'path';
+import { DebugProtocol } from 'vscode-debugprotocol';
+import { LoadedSourceCallFrame } from './callFrame';
+import { functionDescription } from './callFramePresentation';
+
+/** The clients can requests the stack traces frames descriptions in different formats.
+ * We use this function to create the description for the call frame according to the parameters supplied by the client.
+ */
+export function formatCallFrameDescription(callFrame: LoadedSourceCallFrame, formatArgs?: DebugProtocol.StackFrameFormat): string {
```
roblourens: Should this be a method on LoadedSourceCallFrame?

src/chrome/internal/stackTraces/callFrame.ts
```diff
@@ -0,0 +1,128 @@
...
+    readonly codeFlow: CodeFlowFrame<TResource>;
+    readonly scopeChain: Scope[];
+    readonly frameThis?: CDTP.Runtime.RemoteObject;
+    readonly returnValue?: CDTP.Runtime.RemoteObject;
+}
+
+export type CallFrame<TResource extends ScriptOrLoadedSource> =
+    TResource extends ILoadedSource ? LoadedSourceCallFrame :
+    TResource extends IScript ? ScriptCallFrame :
+    ICallFrame<never>; // TODO: Figure out how to change this for never
```
rakatyal: What does the "never" scenario imply?

src/chrome/internal/stackTraces/callFrame.ts
```diff
@@ -0,0 +1,128 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+ import { Location } from '../locations/location';
```
rakatyal: <Nit> Extra white space

src/chrome/internal/stackTraces/callFrame.ts
```diff
@@ -0,0 +1,128 @@
...
+    public get functionName(): string {
+        return this.codeFlow.functionName;
+    }
+}
+
+export class ScriptCallFrame extends BaseCallFrame<IScript> {
+    constructor(
+        public readonly codeFlow: CodeFlowFrame<IScript>,
+        public readonly scopeChain: Scope[],
+        public readonly frameThis: CDTP.Runtime.RemoteObject,
```
rakatyal: What does this variable represent?

src/chrome/internal/stackTraces/callFrame.ts
digeff: never is a value that never happens. I'm using it as an "error condition". If you try to use a variable/value that is typed never for anything, the compiler will complain.

More info at: https://blog.mariusschulz.com/2016/11/18/typescript-2-0-the-never-type

src/chrome/internal/stackTraces/callFrame.ts
digeff: Removed

src/chrome/internal/stackTraces/formatCallFrameDescription.ts
digeff: Moved to CallFramePresentation

PR #392, opened 2019-01-18
Minor fixes and updates

digeff:
Minor fixes and updates

undefinedOnFailure: Scope locations from CDTP sometimes are on line -1 when the doc says it's zero based. Rather than failing we use this to assign undefined location to the scope
isEquivalentTo: This is an equals method. We use this for example in the Breakpoints Delta logic to detect which breakpoints are new and which ones aren't.
breakWhileDebugging: We use this to easily stop the debugger when something "suspicious" happens (some exceptions are expected, so pausing on handle exceptions doesn't always work)
pathToRegex: CDTP doesn't allow us to set a breakpoint in the same location twice. Removing this restriction makes the logic easier for breakpoints because we just add the new breakpoints and remove the old ones.
adaptToSinglIntoToMulti: We have some functions for CDTP and VSCode that map a single stackFrame, or SourceTree to the internal protocol. This method creates an "array map" version of the same function


src/utils.ts
```diff
@@ -522,6 +522,11 @@ export function pathToRegex(aPath: string): string {
         aPath = escapeRegexSpecialChars(fileUrlPrefix) + aPath;
     }
 
+    if (guid) {
+        // Add a guid to the regexp to make it unique, without modifying what it matches. This will allow us to add duplicated breakpoints using CDTP
+        aPath = aPath + `(?:${guid}){0}`;
```
roblourens: No idea what you mean, this does modify what it matches

roblourens: I guess to set a breakpoint for the same path multiple times, but what do we need to do that for?

src/utils.ts
```diff
@@ -655,3 +660,19 @@ export function fillErrorDetails(properties: IExecutionResultTelemetryProperties
         properties.exceptionId = e.id.toString();
     }
 }
+
+export function makeUnique<T>(elements: T[]): T[] {
+    return Array.from(new Set(elements));
+}
+
+export function adaptToSinglIntoToMulti<T, R>(thisObject: object, toSingle: (single: T) => R): (multi: T[]) => R[] {
```
roblourens: `Single`

roblourens: Can you add a comment for what this does? I don't really follow the intent. It's a helper for `map`?

src/chrome/internal/stackTraces/callFrame.ts
```diff
@@ -44,7 +46,8 @@ export class CodeFlowFrame<TResource extends ScriptOrLoadedSource> {
     }
 }
 
-/** This interface represents both the code flow and the state of a call frame.
+/**
```
roblourens: Thanks! Much nicer to my eyes

src/utils.ts
digeff: The regexp should match the same things as before.
Then we can add all the new breakpoints, and then remove all the old ones.

If we don't support 2 breakpoints in the same locations, that code will fail if you change the condition of a breakpoint

src/utils.ts
digeff: /**
 * adaptToSinglIntoToMulti transforms a function that takes a single parameter, to a version that takes an array and applies the function to each parameter.
 * Some usage examples:
 *   public readonly getBPsFromIDs = adaptToSinglIntoToMulti(this, this.getBPFromID);
 *   public readonly toStackFrames = asyncIStackTracePresentationRow(this, this.toStackFrame);
 *   public readonly toSourceTrees = asyncIStackTracePresentationRow(this, this.toSourceTree);
 *   public readonly toBPRecipiesStatus = asyncIStackTracePresentationRow(this, this.toBPRecipieStatus);
 */

src/chrome/internal/stackTraces/callFrame.ts
digeff: My pleasure

src/chrome/internal/scripts/script.ts
```diff
@@ -9,23 +13,25 @@ import { printArray } from '../../collections/printing';
 import { ISourcesMapper } from './sourcesMapper';
 import { IResourceIdentifier, IResourceLocation, newResourceIdentifierMap, parseResourceIdentifier, ResourceName } from '../sources/resourceIdentifier';
 import { IExecutionContext } from './executionContext';
+import { IEquivalenceComparable } from '../../utils/equivalence';
 
-/** This interface represents a piece of code that is being executed in the debugee. Usually a script matches to a file or a url, but that is not always the case.
+/**
+ * This interface represents a piece of code that is being executed in the debugee. Usually a script matches to a file or a url, but that is not always the case.
```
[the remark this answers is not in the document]
digeff: Fixed

src/utils.ts
```diff
@@ -655,3 +660,27 @@ export function fillErrorDetails(properties: IExecutionResultTelemetryProperties
...
+
+/**
+ * adaptToSingleIntoToMulti transforms a function that takes a single parameter, to a version that takes an array and applies the function to each parameter.
+ * Some usage examples:
+ *   public readonly getBPsFromIDs = adaptToSingleIntoToMulti(this, this.getBPFromID);
+ *   public readonly toStackFrames = asyncAdaptToSingleIntoToMulti(this, this.toStackFrame);
+ *   public readonly toSourceTrees = asyncAdaptToSingleIntoToMulti(this, this.toSourceTree);
+ *   public readonly toBPRecipiesStatus = asyncAdaptToSingleIntoToMulti(this, this.toBPRecipieStatus);
+ */
+export function adaptToSingleIntoToMulti<T, R>(thisObject: object, toSingle: (single: T) => R): (multi: T[]) => R[] {
```
roblourens: This is confusing, I don't understand why a caller wouldn't just call map themselves. I don't think this makes code simpler or easier to follow.

digeff: I agree it's simpler and "easier to follow" to write the full method code each time. The trade-off I'm doing here is to avoid writing the same code over and over again, which I think is more error-prone than just writing the code once, and reusing it in this way.

roblourens: What does it eliminate though, one call to 'map'? To me it doesn't seem worth the mental load of introducing a new concept just to save a few characters.

digeff: Currently the code is:
```
    public readonly toStackFrames = asyncAdaptToSingleIntoToMulti(this, this.toStackFrame);
    public readonly toSourceTrees = asyncAdaptToSingleIntoToMulti(this, this.toSourceTree);
    public readonly toBPRecipiesStatus = asyncAdaptToSingleIntoToMulti(this, this.toBPRecipieStatus);

```
Without that the code would be:
```
    public toStackFrames(rows: IStackTracePresentationRow[]): Promise<DebugProtocol.StackFrame[]> {
        return asyncMap(rows, row => this.toStackFrame(row));
    }

```
and the same code for the other 2 methods...

roblourens: Honestly that looks perfectly fine to me. It's one less concept for someone to understand. I would probably skip the helper and just write 

`const frames = rows.map(r => this.toStackFrame(r))` whenever I need to do this conversion, and that doesn't seem like a big deal to me, it's totally transparent.

But it's not a big deal, I won't hold up the PR on it.

src/chrome/collections/printting.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------
```
rakatyal: Nit: Printing.ts

src/chrome/collections/utilities.ts
```diff
@@ -15,3 +19,11 @@ export function determineOrderingOfStrings(left: string, right: string): number
         return 0;
     }
 }
+
+export function singleOne<T>(array: ReadonlyArray<T>): T {
```
rakatyal: Nit: Consider renaming this function to something more descriptive (something of sorts singleElementArrayValue() ? )

digeff: Renamed to: singleElementOfArray

src/chrome/collections/printting.ts
digeff: Deleted extra file

PR #398, opened 2019-02-27
Update script model to support the multiplicities that can happen in production

digeff:
Update script model to support the multiplicities that can happen in production

The previous script model assumed that a lot of multiplicities was 1 to 1 (.e.g: a Loaded source used to be related to a single script).

We can find more complex cases in production, so we updated the models to support the cases that could actually happen in production:

```
/**
 * Multiplicity:
 *   Scripts N [HTMLFile or MultipleTimesLoaded] ... 1 RuntimeSource(LoadedSource)(URLs)
 *   RuntimeSource(LoadedSource)(URLs) N ... 1 DevelopmentSource(LoadedSource)
 *   DevelopmentSource(LoadedSource) N ... M MappedSource(LoadedSource)
 *
 * --- Details ---
 * Scripts N [HTMLFile or MultipleTimesLoaded] ... 1 RuntimeSource(LoadedSource)(URLs)
 *   RuntimeSource(LoadedSource)(URLs) can have N Scripts if it's an .html file with multiple scripts or multiple event handlers
 *   RuntimeSource(LoadedSource)(URLs) can have N Scripts if the same script was loaded multiple times (We've seen this happen in Node when the require cache is deleted)
 *
 * RuntimeSource(LoadedSource)(URLs) N ... 1 DevelopmentSource(LoadedSource)
 * DevelopmentSource(LoadedSource) can be associated with multiple RuntimeSource(LoadedSource)(URLs) if the web-server severs the same file from multiple URLs
 *
 * DevelopmentSource(LoadedSource) N ... M MappedSource(LoadedSource)
 * DevelopmentSource(LoadedSource) can be associated with multiple MappedSource(LoadedSource) if files were bundled or compiled with TypeScript bundling option
 * MappedSource(LoadedSource) can be associated with multiple DevelopmentSource(LoadedSource) if the same typescript file gets bundled into different javascript files
 *
 * Additional notes:
 * It's extremelly unlikely, but it's possible for a .js file to be the MappedSource of a Script A, the RuntimeSource of a different script B, and the DevelopmentSource for a different script C at the same time
 */
```

### DIFF
--- a/src/chrome/internal/scripts/sourcesMapper.ts
+++ b/src/chrome/internal/scripts/sourcesMapper.ts
@@ -2,61 +2,133 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
-import { LineNumber, ColumnNumber, createColumnNumber, createLineNumber } from '../locations/subtypes';
-import { SourceMap } from '../../../sourceMaps/sourceMap';
+import { createColumnNumber, createLineNumber } from '../locations/subtypes';
+import { SourceMap, SourceRange } from '../../../sourceMaps/sourceMap';
+import { LocationInLoadedSource, LocationInScript, Position } from '../locations/location';
+import { ILoadedSource } from '../sources/loadedSource';
+import { IResourceIdentifier, parseResourceIdentifier } from '../sources/resourceIdentifier';
+import { IScript } from './script';
 
-export interface ISourcesMapper {
+export interface ISourceToScriptMapper {
+    getPositionInScript(positionInSource: LocationInLoadedSource): LocationInScript | null;
+}
+
+export interface IScriptToSourceMapper {
+    getPositionInSource(positionInScript: LocationInScript): LocationInLoadedSource | null;
+}
+
+export interface ISourceMapper extends ISourceToScriptMapper, IScriptToSourceMapper { }
+
+export interface IMappedSourcesMapper extends ISourceMapper {
     readonly sources: string[];
-    getPositionInSource(positionInScript: IPositionInScript): IPositionInSource | null;
-    getPositionInScript(positionInSource: IPositionInSource): IPositionInScript | null;
-}
-
-interface IPositionInSource {
-    readonly source: string;
-    readonly line: LineNumber;
-    readonly column?: ColumnNumber;
-}
-
-interface IPositionInScript {
-    readonly line: LineNumber;
-    readonly column?: ColumnNumber;
 }
 
 /** This class maps locations from a script into the sources form which it was compiled, and back. */
-export class SourcesMapper implements ISourcesMapper {
-    public getPositionInSource(positionInScript: IPositionInScript): IPositionInSource | null {
-        const mappedPosition = this._sourceMap.authoredPositionFor(positionInScript.line, positionInScript.column || 0);
-        return mappedPosition && mappedPosition.source && mappedPosition.line
-            ? { source: mappedPosition.source, line: createLineNumber(mappedPosition.line), column: createColumnNumber(mappedPosition.column) }
-            : null;
+export class MappedSourcesMapper implements IMappedSourcesMapper {
+    private readonly _rangeInSources: Map<IResourceIdentifier, SourceRange>;
+
+    public getPositionInSource(positionInScript: LocationInScript): LocationInLoadedSource | null {
+        const scriptPositionInResource = this._script.rangeInSource.start.position;
+
+        // All the lines need to be adjusted by the relative position of the script in the resource (in an .html if the script starts in line 20, the first line is 20 rather than 0)
+        const lineNumberRelativeToScript = positionInScript.position.lineNumber - scriptPositionInResource.lineNumber;
+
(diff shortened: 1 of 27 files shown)


src/chrome/internal/scripts/sourcesMapper.ts
```diff
@@ -2,61 +2,133 @@
...
+
+        // All the lines need to be adjusted by the relative position of the script in the resource (in an .html if the script starts in line 20, the first line is 20 rather than 0)
+        const lineNumberRelativeToScript = positionInScript.position.lineNumber - scriptPositionInResource.lineNumber;
+
+        // The columns on the first line need to be adjusted. Columns on all other lines don't need any adjustment.
+        const columnNumberRelativeToScript = (lineNumberRelativeToScript === 0 ? scriptPositionInResource.columnNumber : 0) + (positionInScript.position.columnNumber || 0);
+
+        const mappedPosition = this._sourceMap.authoredPositionFor(lineNumberRelativeToScript, columnNumberRelativeToScript);
+
+        if (mappedPosition && mappedPosition.source && mappedPosition.line) {
```
roblourens: `mappedPosition.line` could be 0 right, dangerous truthiness check?

src/chrome/internal/sources/identifiedLoadedSource.ts
```diff
@@ -0,0 +1,78 @@
+import * as fs from 'fs';
+import { IScript } from '../scripts/script';
+import { IResourceIdentifier } from './resourceIdentifier';
+import { ILoadedSource, IScriptMapper, ICurrentScriptRelationshipsProvider as IScriptMapperProvider, ContentsLocation, SourceScriptRelationship, ImplementsLoadedSource, ScriptAndSourceMapper } from './loadedSource';
+import { ILoadedSourceToScriptRelationship } from './loadedSourceToScriptRelationship';
+import _ = require('lodash');
```
roblourens: It looks like there are types for lodash at `@types/lodash`

roblourens: test

roblourens: foo
bar

src/chrome/internal/sources/identifiedLoadedSource.ts
```diff
@@ -0,0 +1,78 @@
...
+import { printArray } from '../../collections/printing';
+import { LocationInScript, LocationInLoadedSource } from '../locations/location';
+
+/**
+ * Loaded Source classification:
+ * Is the script content available on a single place, or two places? (e.g.: You can find similar scripts in multiple different paths)
+ *  1. Single: Is the single place on storage, or is this a dynamic script?
+ *      Single path on storage: RuntimeScriptRunFromStorage
```
roblourens: Let's talk about this, I can't tell what happens in the case where there is a `/foo.js` script in the runtime, and also a file at that path on disk, but the script in the runtime is dynamic and has different contents than the file on disk which is actually the source. This is anything that transpiles in memory like ts-node.

src/chrome/internal/sources/identifiedLoadedSource.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public toString(): string {
+        return `src:${this.identifier}`;
+    }
+
+    public static create<TString extends string>(identifier: IResourceIdentifier<TString>, sourceScriptRelationship: SourceScriptRelationship,
+        currentScriptRelationshipsProvider: IScriptMapperProvider): IdentifiedLoadedSource<TString> {
+
+        const contentsLocation = fs.existsSync(identifier.textRepresentation) ? ContentsLocation.PersistentStorage : ContentsLocation.DynamicMemory;
```
roblourens: Could probably make this async

src/chrome/internal/sources/identifiedLoadedSource.ts
roblourens: Also there is a difference between "Dynamic" scripts that have paths and ones that don't

src/chrome/internal/scripts/sourcesMapper.ts
digeff: Nice catch

src/chrome/internal/sources/identifiedLoadedSource.ts
digeff: As discussed offline, the current code won't support that case nor ts-node. We can implement it in a future version.

src/chrome/internal/sources/identifiedLoadedSource.ts
digeff: I'll add a TODO to make that change.

The method is indirectly called from the Script class constructor. I tried to make it async at certain point, but I couldn't figure out how to make the Script class constructor be async (and if I want to make the class immutable, I cannot modify the member variables after the constructor finishes).

src/chrome/internal/sources/identifiedLoadedSource.ts
digeff: Changed

PR #403, opened 2019-03-08
Client Model

digeff:
gulpfile updated to v4
ConnectedCDAConfiguration this class is meant to have all the configuration that we use at runtime in a single place
ChromeDebugAdapter (v2): New version of the debug adapter. It's split into three states: Uninitialized, Unconnected and Connected.
ConnectedCDA: Handles the requests when we are connected to Chrome. The components implement the ICommandHandlerDeclarer interface to register requests they support here.
UnconnectedCDA: Handles the launch/attach requests, and creates the ConnectedCDA instance that we'll use at runtime (This class/code still needs a lot of work).
UninitializedCDA: Handles the initialize request, and responds the client capabilities.
DelayMessagesUntilInitializedSession: The protocol doesn't support sending events before we send the initialized message. This classes stores the events that are generated before that happens, and sends them after we send the initialized message.
DoNotPauseWhileSteppingSession: This classes ensures that we'll wait until step requests finish before we send a stopped event to the client.
BidirectionalHandles: Class to store the handles/ids that we send in the vs code protocol messages
IDebuggeeLauncher/IDebuggeeRunner: Classes that each debugger must implement to tell -core how to launch and run the debuggee
DependencyInjection: We use this class to hide the actual DI framework that we use, in case we decide to change to a different framework later
IExtensibilityPoints: In v2 each debugger will use the extensibility points to add, or modify the behavior of -core for it's particular debugger. (This class/code is still a work in progress)
Logging: We plan to use this service instead of the global logger. service.
MethodsCalledLogger: Proxy Wrapper to automatically log the calls to an object. We wrap all the DependencyInjection components with these, to log all the calls that are made between each component. (This class/code is still a work in progress)
PrintObjectDescription: This class figures out the best way to print an object (This is useful when the object doesn't have a custom .toString() implementation) (This class/code is still a work in progress)

### DIFF
--- a/src/chrome/client/doNotPauseWhileSteppingSession.ts
+++ b/src/chrome/client/doNotPauseWhileSteppingSession.ts
@@ -1,0 +1,54 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as utils from '../../utils';
+import { BaseWrappedSession } from './session';
+import { DebugProtocol } from 'vscode-debugprotocol';
+
+const steppingRequests = {
+    continue: true,
+    next: true,
+    stepIn: true,
+    stepOut: true,
+    pause: true,
+    restartFrame: true,
+};
+
+export class DoNotPauseWhileSteppingSession extends BaseWrappedSession {
+    private readonly _onFlightSteppingRequests = new Set<Promise<void>>();
+
+    public async dispatchRequest(request: DebugProtocol.Request): Promise<void> {
+        const response = this._wrappedSession.dispatchRequest(request);
+        if (this.isSteppingRequest(request)) {
+            // We track the on-flight stepping requests
+            this._onFlightSteppingRequests.add(response);
+            const finallyHandler = () => { this._onFlightSteppingRequests.delete(response); };
+            return response.then(finallyHandler, finallyHandler);
+        } else {
+            return await response;
+        }
+    }
+
+    public async sendEvent(event: DebugProtocol.Event): Promise<void> {
+        if (event.event === 'stopped') {
+            // If this is a stopped event, we try to wait until there are no stepping requests in flight, or we timeout
+            await utils.promiseTimeout(this.waitUntilThereAreNoOnFlightSteppingRequests(), /*timeoutMs=*/300);
+        }
+
+        this._wrappedSession.sendEvent(event);
+    }
+
+    private isSteppingRequest(request: DebugProtocol.Request): boolean {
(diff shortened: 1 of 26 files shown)


src/chrome/client/doNotPauseWhileSteppingSession.ts
```diff
@@ -0,0 +1,54 @@
...
+const steppingRequests = {
+    continue: true,
+    next: true,
+    stepIn: true,
+    stepOut: true,
+    pause: true,
+    restartFrame: true,
+};
+
+export class DoNotPauseWhileSteppingSession extends BaseWrappedSession {
```
roblourens: I don't quite understand why this is needed, if you are stepping quickly, don't you want to see the 'stopped' event ASAP, rather than waiting until stepping is done?

src/chrome/logging/methodsCalledLogger.ts
```diff
@@ -0,0 +1,103 @@
...
+                            throw exception;
+                        }
+                    };
+                } else {
+                    return originalPropertyValue;
+                }
+            }
+        };
+
+        return new Proxy<T>(this._objectToWrap, handler);
```
roblourens: One possible concern with this is that debugging with Proxies is harder because you don't know what actual object is behind the proxy. Is this going to wrap every object, or how often is it going to be used?

src/chrome/client/doNotPauseWhileSteppingSession.ts
digeff: This feature already existed on v1 here: https://github.com/Microsoft/vscode-chrome-debug-core/blob/ae8afa131a1dc37008d6a7df1869956ecc1c5d76/src/chrome/chromeDebugAdapter.ts#L731
I refactored/rewrote it so it's not mixed with the Pause logic.
I assumed that things broke if you sent a pause before sending a response to the step request that created the pause...

src/chrome/logging/methodsCalledLogger.ts
digeff: At the moment it's wrapping all the objects that the Dependency Injection framework injects.
I'm also planning on adding an option to turn off this logging, so there'll be a way to avoid having the objects being wrapped in the proxies. Does that alleviate your concern?

src/chrome/client/doNotPauseWhileSteppingSession.ts
roblourens: I think pinezorro cares about that, vscode doesn't.

src/chrome/logging/methodsCalledLogger.ts
roblourens: Yeah I think so

PR #406, opened 2019-03-13
Set breakpoints model

digeff:
BPRecipeAtLoadedSourceLogic: Handles setting breakpoints on sources associated with scripts that are already loaded
BPRsDeltaCalculator: Calculates the difference between two sets of breakpoint recipes, to find out which breakpoints were added, and removed
BPRecipeStatusCalculator: Calculats the current status of a BPRecipe (Bound vs Unbound)
BreakpointsEventSystem: Mini-event system to hide how the events work from the breakpoint's logic
BreakpointsUpdater: Handles updating the breakpoints for a particular source
ExistingBPsForJustParsedScriptSetter: Handles adding the neccesary breakpoints to a script that was just parsed
HitCountBreakpoints: Handles hit count breakpoint (This is not currently working/being used)
PauseScriptLoadsToSetBPs: Implements the break-on-load feature. When we load a script, we pause on it until we've set all the breakpoints for that script
BreakpointsSetForScriptFinder : Finds the breakpoints that were set for a particular script
CurrentBPRecipesForSourceRegistry : Stores the BP Recipies currently set for a source
DebuggeeBPRsSetForClientBPRFinder : Finds all the debuggee BPRs that were set for a client BPR
ExecutionLogger: Class that the event dispatcher framework uses for logging (We'll probably refactor/delete this)
Communicator : Main class of the event dispatcher component
NamespaceReverseLookupCreator: Utility used for the communicator to generate better error messages

Added a small component to publish typed events used in the set breakpoints feature (This component is way overkill for what I'm using it right now. Initially I was using that component instead of the DependencyInjection framework so it supports much more many things that what we are using it for atm. Given that it's already working, I'm leaving it as-is for the time being).

### DIFF
--- a/src/chrome/internal/breakpoints/features/bpRecipeAtLoadedSourceLogic.ts
+++ b/src/chrome/internal/breakpoints/features/bpRecipeAtLoadedSourceLogic.ts
@@ -1,0 +1,141 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as _ from 'lodash';
+import * as chromeUtils from '../../../chromeUtils';
+import { IDebuggeeBreakpointsSetter } from '../../../cdtpDebuggee/features/cdtpDebuggeeBreakpointsSetter';
+import { IBreakpointFeaturesSupport } from '../../../cdtpDebuggee/features/cdtpBreakpointFeaturesSupport';
+import { IEventsToClientReporter } from '../../../client/eventsToClientReporter';
+import { ReasonType } from '../../../stoppedEvent';
+import { CDTPBreakpoint } from '../../../cdtpDebuggee/cdtpPrimitives';
+import { DebuggeeBPRsSetForClientBPRFinder } from '../registries/debuggeeBPRsSetForClientBPRFinder';
+import { BPRecipeInLoadedSource } from '../BaseMappedBPRecipe';
+import { ConditionalPause, AlwaysPause } from '../bpActionWhenHit';
+import { PausedEvent } from '../../../cdtpDebuggee/eventsProviders/cdtpDebuggeeExecutionEventsProvider';
+import { BPRecipe } from '../bpRecipe';
+import { ISource } from '../../sources/source';
+import { LocationInScript, Position } from '../../locations/location';
+import { createColumnNumber, createLineNumber } from '../../locations/subtypes';
+import { RangeInResource } from '../../locations/rangeInScript';
+import { logger } from 'vscode-debugadapter/lib/logger';
+import { asyncMap } from '../../../collections/async';
+import { wrapWithMethodLogger } from '../../../logging/methodsCalledLogger';
+import { BaseNotifyClientOfPause, IActionToTakeWhenPaused, NoActionIsNeededForThisPause } from '../../features/actionToTakeWhenPaused';
+import { IDebuggeePausedHandler } from '../../features/debuggeePausedHandler';
+import { IBreakpointsEventsPublisher } from './BreakpointsEventSystem';
+import { BPRecipeIsUnbound } from '../bpRecipeStatusForRuntimeLocation';
+
+export class HitBreakpoint extends BaseNotifyClientOfPause {
+    protected reason: ReasonType = 'breakpoint';
+
+    constructor(protected readonly _eventsToClientReporter: IEventsToClientReporter) {
+        super();
+    }
+}
+
+export interface IBreakpointsInLoadedSource {
+    addBreakpointAtLoadedSource(bpRecipe: BPRecipeInLoadedSource<ConditionalPause | AlwaysPause>): Promise<CDTPBreakpoint[]>;
+    removeDebuggeeBPRs(clientBPRecipe: BPRecipe<ISource>): Promise<void>;
+}
+
+/**
+ * Handles setting breakpoints on sources that are associated with scripts already loaded
+ */
+export class BPRecipeAtLoadedSourceLogic implements IBreakpointsInLoadedSource {
+    private readonly doesTargetSupportColumnBreakpointsCached: Promise<boolean>;
+    private readonly _publishDebuggeeBPRecipeAdded = this._breakpointsEventsPublisher.publisherForDebuggeeBPRecipeAdded();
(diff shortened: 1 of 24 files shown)


src/chrome/internal/breakpoints/features/bpRecipeAtLoadedSourceLogic.ts
```diff
@@ -0,0 +1,141 @@
...
+                let breakpoints: CDTPBreakpoint[];
+                if (!runtimeSource.doesScriptHasUrl()) {
+                    breakpoints = [await this._targetBreakpoints.setBreakpoint(bpRecipeInBestLocation)];
+                } else {
+                    /**
+                     * If the script is a local file path, we *need* to transform it into an url to be able to set the breakpoint
+                     *
+                     * If the script has an URL and it's not a local file path, then we could actually leave it as-is.
+                     * We transform it into a regexp anyway to add a GUID to it, so CDTP will let us add the same breakpoint/recipe two times (using different guids).
+                     * That way we can always add the new breakpoints for a file, before removing the old ones (except if the script doesn't have an URL)
```
roblourens: I thought that with the breakpoint diffing, we wouldn't have to remove then immediately re-add new BPs?

src/chrome/internal/breakpoints/features/breakpointsEvents.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { NotificationChannelIdentifier } from '../../../communication/notificationsCommunicator';
+import { registerChannels } from '../../../communication/channel';
+import { BPRecipeInSource } from '../bpRecipeInSource';
+import { CDTPBPRecipe, CDTPBreakpoint } from '../../../cdtpDebuggee/cdtpPrimitives';
+import { BPRecipeIsUnbound } from '../bpRecipeStatusForRuntimeLocation';
+
+const _breakpointsEvents = {
```
roblourens: I don't really understand this stuff, would this be replaced by DI in theory in the future or is this something else?

src/chrome/internal/breakpoints/features/bpRecipeAtLoadedSourceLogic.ts
digeff: We have several cases. For "normal" breakpoints that are set in a script which has a url we are adding all the new breakpoints first, and removing all the breakpoints that we don't need any more second. We can only do that, because we are adding the random guid to each breakpoint.

If we didn't add a random guid to each breakpoint, we wouldn't be able to add new breakpoint first, and remove old breakpoints later because Chrome considers a breakpoint with and without a condition to be the same breakpoint. If you go to VS Code, and you add a breakpoint, and then you right click to add a condition to it, you are swapping a breakpoint for a conditional breakpoint in that same line. If you tried to add that conditional breakpoint in chrome without using a guid, i'd complain that a breakpoint on that line already exists, so your only option would be to remove the normal breakpoint first, and then add the conditional breakpoint, thus we'd have a race condition. Because we are using the random guid for the breakpoints, the new conditional breakpoint is seen as a different breakpoint than the normal breakpoint, and we can add it before removing the old normal breakpoint.

For scripts without an url, we cannot use a regexp. We cannot add new breakpoints first, and remove old breakpoints later because if a user adds a breakpoint, and then converts it  to a conditional breakpoint, chrome will give us an error, so in that scenario we need to remove the breakpoints first, and add breakpoints later.

The other reason we need the guid, is for a part of the breakpoints logic that we haven't implemented yet. We need to add a component to BreakpointsUpdater.updateBreakpointsForFile on line 100 so when we cannot resolve a source because it's associated script hasn't been loaded, we should set the breakpoint in the project file path to see if it works e.g.: C:\myproject\myap\app.js. That breakpoint will only work for Node.js applications and/or static web applications hosted on storage for Chrome. In those cases this new component that we'll add will add a breakpoint to C:\myproject\myap\app.js, and after the script is loaded, ExistingBPsForJustParsedScriptSetter and BPRecipeAtLoadedSourceLogic will try to re-add the same breakpoint. Having the guid simplifies that logic, because we can just add the breakpoint twice (once by each component), and then let the new component that sets breakpoints before loading remove that breakpoint.

src/chrome/internal/breakpoints/features/breakpointsEvents.ts
digeff: My current plan is to try to use the DI framework with different components that are used from different layers in the application, and we want to hide the exact implementation, and also for components that might need to be overwritten by chrome-debug, or node-debug.

These events are used to communicate between different sub-components of the breakpoint feature, which are more coupled with each other than all the components we use in the DI framework.

In summary my plan is to:
Use the DI framework for composing different components from different layers
Use this event dispatcher code to communicate inside a single "logical" component where the sub-components are more coupled.

I'm happy to talk more about this, if this approach doesn't make sense to you.

src/chrome/internal/breakpoints/features/bpRecipeAtLoadedSourceLogic.ts
roblourens: What about the diffing though? I must be missing something because the response doesn't seem to answer my question.

> For scripts without an url, we cannot use a regexp

Why not, aren't we doing that currently?

src/chrome/internal/breakpoints/features/breakpointsEvents.ts
roblourens: It sounds like this stuff fills the same purpose as DI, just within one feature area instead of globally. Maybe I don't really understand how it works but it seems like it would be simpler to just use the same DI infrastructure. If you don't want all the detailed breakpoints components to be accessible globally, other DI systems I've seen have some sort of concept of a DI "container" that you can register in instead of registering them globally. Then it would be like Breakpoints owns its own private set of injectable components without introducing new terms and concepts for someone to learn.

roblourens: I don't know whether inversify has something like that

digeff: It seems inversify has something that might work: https://github.com/inversify/InversifyJS/blob/master/wiki/hierarchical_di.md
Is it okay if I investigate this, and do this change after we merge this PR?

roblourens: Sure

PR #409, opened 2019-03-14
Fixes to existing v1 files

digeff:
- Fixing the format of some files
- Adding _ to unused variables
- Rename CDP to CDTP
- Add types and | undefineds to eventually enable TypeScript strict mode
- Remove references to CRDPMultiplexor
- Use version library instead of custom version code
- Adapt code to the v2 model
- Add I prefix to interfaces
- Adapt transformers to work with the v2 model


src/transformers/eagerSourceMapTransformer.ts
```diff
@@ -44,10 +48,10 @@ export class EagerSourceMapTransformer extends BaseSourceMapTransformer {
 
     private discoverSourceMapForGeneratedScript(generatedScriptPath: string): Promise<void | SourceMap> {
         return this.findSourceMapUrlInFile(generatedScriptPath)
-            .then(uri => {
+            .then(async uri => {
                 if (uri) {
                     logger.log(`SourceMaps: sourcemap url parsed from end of generated content: ${uri}`);
-                    return this._sourceMaps.processNewSourceMap(generatedScriptPath, uri, this._isVSClient);
+                    await this._sourceMaps.processNewSourceMap(generatedScriptPath, uri, this._isVSClient);
```
roblourens: It looks like this wants to return the SourceMap?

digeff: We are currently not using this result value. I changed this method to return Promise<void> so this won't be confusing

PR #414, opened 2019-03-20
Fix bug

digeff:
Fix stacktrace bug


src/chrome/internal/stackTraces/stackTracePresenter.ts
```diff
@@ -76,8 +76,9 @@ export class StackTracePresenter implements IInstallableComponent {
         return [/* Description of this section of async frames */<IStackTracePresentationRow>new StackTraceLabel(stackTrace.description)].concat(thisSectionAsyncFrames, parentAsyncFrames);
     }
 
-    private framesRange(allStackFrames: IStackTracePresentationRow[], firstFrameIndex: number, framesCount?: number) {
-        return allStackFrames.slice(firstFrameIndex, firstFrameIndex + framesCount);
+    private framesRange(allStackFrames: IStackTracePresentationRow[], firstFrameIndex: number, framesCountOrNull?: number) {
+        const framesCount = framesCountOrNull !== null ? firstFrameIndex + framesCountOrNull : allStackFrames.length - firstFrameIndex;
```
roblourens: Since the argument is `?` it will be undefined not null

digeff: Nice catch :)
I'm planning on working on enabling strict mode soon to avoid these issues.
Fixing the type to show that we are actually passing null around

PR #417, opened 2019-03-29
Enable strict null type checking

digeff:
Enabled strict null checking for Typescript.
Modified all the code necessary to make the project compile with that option.
Also deleted some code from the transformers that wasn't being used any more.

### DIFF
--- a/src/chrome/logging/printObjectDescription.ts
+++ b/src/chrome/logging/printObjectDescription.ts
@@ -14,16 +14,23 @@ export function printObjectDescription(objectToPrint: unknown, fallbackPrintDesc
             // This is a noice-json-rpc proxy
             printed = 'CDTP Proxy';
         } else {
-            const toString = objectToPrint.toString();
-            if (toString !== '[object Object]') {
-                printed = toString;
-            } else if (isJSONObject(objectToPrint)) {
-                printed = JSON.stringify(objectToPrint);
-            } else if (objectToPrint.constructor === Object) {
-                printed = fallbackPrintDescription(objectToPrint);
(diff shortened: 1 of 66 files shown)


src/chrome/logging/printObjectDescription.ts
```diff
@@ -14,16 +14,23 @@ export function printObjectDescription(objectToPrint: unknown, fallbackPrintDesc
...
+                if (toString !== '[object Object]') {
+                    printed = toString;
+                } else if (isJSONObject(objectToPrint)) {
+                    printed = JSON.stringify(objectToPrint);
+                } else if (objectToPrint.constructor === Object) {
+                    printed = fallbackPrintDescription(objectToPrint);
+                } else {
+                    printed = `${objectToPrint}(${objectToPrint.constructor.name})`;
+                }
+                }
```
roblourens: indentation

src/sourceMaps/sourceMaps.ts
```diff
@@ -17,59 +17,34 @@ export class SourceMaps {
         this._sourceMapFactory = new SourceMapFactory(pathMapping, sourceMapPathOverrides, enableSourceMapCaching);
     }
 
-    /**
-     * Returns the generated script path for an authored source path
-     * @param pathToSource - The absolute path to the authored file
-     */
-    public getGeneratedPathFromAuthoredPath(authoredPath: string): string {
-        authoredPath = authoredPath.toLowerCase();
```
roblourens: Did this move or is it not needed anymore?

digeff: We replaced this by the Script model and it's developmentSource, runtimeSource, mappedSource, and sourceMapper properties

PR #418, opened 2019-04-01
Fixing disconnect issue

EricCornelson:
Previously, the debug adapter would wait for disconnect to be processed and then send the response. This was problematic because part of the disconnect process was to shut down the connection to the client. This would leave the client waiting for the disconnect response which would never come.

### DIFF
--- a/src/chrome/chromeDebugSession.ts
+++ b/src/chrome/chromeDebugSession.ts
@@ -132,8 +132,15 @@ export class ChromeDebugSession extends LoggingDebugSession implements IObservab
             try {
                 logger.verbose(`From client: ${request.command}(${JSON.stringify(request.arguments) })`);
                 telemetryPropertyCollector.addTelemetryProperty('requestType', request.type);
+                // Disconnect request is a special case--the response needs to be sent before
+                // we do our "shutdown" processing.
+                if (request.command === 'disconnect') {
+                    this.sendResponse(response);
+                }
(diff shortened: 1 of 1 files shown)


src/chrome/chromeDebugSession.ts
```diff
@@ -132,8 +132,15 @@ export class ChromeDebugSession extends LoggingDebugSession implements IObservab
...
+                if (request.command === 'disconnect') {
+                    this.sendResponse(response);
+                }
                 response.body = await this._debugAdapter.processRequest(<CommandText>request.command, request.arguments, telemetryPropertyCollector);
-                this.sendResponse(response);
+                if (request.command !== 'disconnect') {
+                    this.sendResponse(response);
+                }
             } catch (e) {
                 if (!this.isEvaluateRequest(request.command, e)) {
```
roblourens: If disconnect fails, we could end up here and send the response twice

EricCornelson: Which line are you talking about? We only send the response in the second block (line 142) if the incoming request was anything other than a disconnect.

roblourens: The lines inside the `catch`. If handling the request fails, `processRequest` can throw.

roblourens: Yeah, it's not likely that it would actually fail, and if it does fail, vscode is probably still going to shut down the DA anyway. I think we should just make sure we log the issue and avoid sending a fail response after we already sent the success response.

PR #420, opened 2019-04-05
Enable strict mode

digeff:
Finished enabling strict mode


src/nullLogger.ts
```diff
@@ -1,13 +1,15 @@
+import { ILogger, ILoggingConfiguration } from './chrome/internal/services/logging';
```
roblourens: Watch out for the imports above the copyright, autoimport will do that...

digeff: Fixed

PR #428, opened 2019-04-18
- Several bug fixes to make the int test pass

digeff:
- Merge changes from core-v2 from v1
- Several fixes to make the int tests work:
1. chromeDebugSession.ts: Removed AvailableCommands to make illegal requests respond a failure
2. sourceToClientConverter.ts: Updated logic so stack traces would return file paths as file paths instead of file:/// urls
3. sourceToScriptMapper.ts: setting a breakpoint on the semi-colon on "x++;" was failing. This searches for a break point in the whole line to make that work
4. syncStepping.ts: After stepping-in the stopped reason was: "debugger_statement" instead of "step". This fixes it.
5. sourceMap.ts: Some of our tests had source mappings where the lines and source were null, and that was crashing that method
6. singleBreakpointSetter.ts: Fixed a race condition while enabling instrumentation for scripts

### DIFF
--- a/src/chrome/collections/validatedSet.ts
+++ b/src/chrome/collections/validatedSet.ts
@@ -3,6 +3,8 @@ import { breakWhileDebugging } from '../../validation';
 
 export interface IValidatedSet<K> extends Set<K> {
     addOrReplaceIfExists(key: K): this;
+    deleteIfExists(key: K): boolean;
+    toArray(): K[];
 }
 
 /** A set that throws exceptions instead of returning error codes. */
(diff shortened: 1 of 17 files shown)


src/chrome/collections/validatedSet.ts
```diff
@@ -39,6 +41,10 @@ export class ValidatedSet<K> implements IValidatedSet<K> {
         return true;
     }
 
+    public deleteIfExists(key: K): boolean {
```
roblourens: You could just call it `delete`?

PR #429, opened 2019-04-23
Add a pipeline task to run integration tests from chrome-debug

EricCornelson:


gulp-test-chrome-debug.js
```diff
@@ -0,0 +1,80 @@
...
+ * Run `npm install` on vscode-chrome-debug
+ */
+function install() {
+    return runInDebug('npm install');
+}
+
+/**
+ * Substitute our version of -core for the tests
+ */
+function setCoreVersion(cb) {
```
roblourens: You could use `npm link` in the -core folder, then `npm link vscode-chrome-debug-core` in the `vscode-chrome-debug` folder

gulp-test-chrome-debug.js
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Run the integration tests for vscode-chrome-debug
+ */
+function intTest() {
+    return runInDebug('npm run allIntTest', {cwd: './vscode-chrome-debug/'});
+}
+
+exports.clean = clean;
```
roblourens: I don't think these need to be exported

gulp-test-chrome-debug.js
EricCornelson: True, but that's just using symlinks to do the same thing, no?

roblourens: Yes, it's probably shorter and more of a standard way to do it.

gulp-test-chrome-debug.js
EricCornelson: I left them exported so I could test each step individually, but it's not required for CI

gulp-test-chrome-debug.js
EricCornelson: Fair. I'll change it.

roblourens:
Looks like the build is broken, is that from this?

EricCornelson:
Yeah, something is failing on the git clone, investigating now

gulp-test-chrome-debug.js
```diff
@@ -0,0 +1,75 @@
+const { series } = require('gulp');
+const del = require('del');
+const run = require('gulp-run-command').default;
+
+const IS_WINDOWS = (process.platform == 'win32');
+
+/**
+ * Run a command in the vscode-chrome-debug working directory
+ */
+function runInDebug(cmd) {
```
digeff: I'd rename this to runInChromeDebug. My mind goes to a very different places with runInDebug...

gulp-test-chrome-debug.js
```diff
@@ -0,0 +1,75 @@
...
+ * Run `npm install` on vscode-chrome-debug
+ */
+function install() {
+    return runInDebug('npm install');
+}
+
+/**
+ * Substitute our version of -core for the tests
+ */
+async function setCoreVersion(cb) {
```
digeff: I'd rename this to setCoreVersion linkChromeDebugToCore
setCoreVersion makes me think that you are setting the version as in 4.5.6

gulp-test-chrome-debug.js
```diff
@@ -1,64 +1,69 @@
 const { series } = require('gulp');
 const del = require('del');
 const run = require('gulp-run-command').default;
+const fs = require('fs');
+const extract = require('extract-zip');
+const path = require('path');
+const request = require('request');
 
-const IS_WINDOWS = (process.platform == 'win32');
+const CHROME_DEBUG_V2_URL = 'https://github.com/Microsoft/vscode-chrome-debug/archive/v2.zip';
```
roblourens: Does this just let you download any branch as a zip? Cool, but why not git clone?

EricCornelson: I tried for a few days to get git-clone working in azure with no luck. The docs seem to say that it's possible with the yaml, but [this issue](https://github.com/Microsoft/azure-pipelines-yaml/issues/103) seems to say otherwise. From there, I've tried within gulp to just launch git with spawn, but something seems to be messed up with the git install and/or path on the build machines, as I consistently get this error:

> git: 'remote-https' is not a git command. See 'git --help'.

Without being able to remote in to a machine somewhere and do some testing (or find a way to set up a VM or something to mirror the setup) I'm finding it really hard to debug the issue. Downloading the zip is a workaround since I don't want to spend a lot more time on this issue. Any tips on  are appreciated, though :)

EricCornelson: Uhg, still failing with the same issue because something inside `npm install` is trying to run a git command

PR #430, opened 2019-04-24
Fix issues to make int tests pass

digeff:
- breakpointsUpdater.ts: Move the breakpoint update logic to setBreakpointsRequestHandler.ts to synchronize the updates with the setBreakpoints request, and prevent a race condition (We shouldn't attempt to send a breakpoint update for a breakpoint before that breakpoint appears on the response of setBreakpoints, and the client knows the id).

- hitCountBreakpointsSetter.ts: Added more logging

- setBreakpointsRequestHandler.ts: Synchronize the breakpoints update, so they'll wait for the setBreakpoints requests that are on flight to finish before sending the update. If we don't do this, we can try to send an update for a breakpoint that hasn't been assigned an id yet, and the this._eventsToClientReporter.sendBPStatusChanged method will fail.

- singleBreakpointSetter.ts: Instead of enabling the DOM Instrumentation for Scripts when we get the first breakpoint, we enable it always at the start, to address race conditions that can happen if we are processing a ScriptParsed event while we enable the DOM Instrumentation for Scripts.

- sourcesMapper.ts: Added more logging, and updated in response to the changes in sourceMap.ts to use Position and IResourceIdentifier instead of strings.

- syncStepping.ts: Created UserPaused and CurrentlyPausing to force the debugger to pause after we send the Pause request, even if the message is a DOM Instrumentation pause (Without this change we would auto-resume if we got a DOM Instrumentation pause after reciving a pause request from the user).

- sourceMap.ts:
	1. Moved the ISourcePathDetails logic into it's own class: SourcePathMappingCalculator.
	2. Updated the class to use ResourceIdentifiers, and to always normalize the sources strings with .canonicalized (The tests were failing because the string we looked up was normalized differently than the one we stored on the map).
	3. Updated authoredPosition, generatedPositionFor, and allGeneratedPositionsForBothDirections to provide a more strict error checking, and to look for mappings in both directions if needed

### DIFF
--- a/src/chrome/utils/promises.ts
+++ b/src/chrome/utils/promises.ts
@@ -1,5 +1,22 @@
+import { asyncMap } from '../collections/async';
+
 /*---------------------------------------------------------
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
 export type PromiseOrNot<T> = Promise<T> | T;
+
+/**
+ * Wait until the promise gets either resolved or rejected. If the promise gets rejected, waitForEnd will ignore that rejection and succeed anyways.
+ */
+export async function waitForEnd(...manyPromisesToWaitFor: Promise<unknown>[]): Promise<void> {
+    await asyncMap(manyPromisesToWaitFor, waitForSinglePromiseToEnd);
+}
+
+async function waitForSinglePromiseToEnd(promiseToWaitFor: Promise<unknown>): Promise<void> {
+    try {
+        await promiseToWaitFor;
+    } catch {
+        // Ignore failures
+    }
+}
--- a/src/chrome/utils/promises.ts
+++ b/src/chrome/utils/promises.ts
@@ -1,22 +1,5 @@
-import { asyncMap } from '../collections/async';
-
 /*---------------------------------------------------------
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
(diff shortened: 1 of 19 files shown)


src/chrome/utils/promises.ts
```diff
@@ -1,5 +1,22 @@
...
  *--------------------------------------------------------*/
 
-export type PromiseOrNot<T> = Promise<T> | T;
\ No newline at end of file
+export type PromiseOrNot<T> = Promise<T> | T;
+
+/**
+ * Wait until the promise gets either resolved or rejected. If the promise gets rejected, waitForEnd will ignore that rejection and succeed anyways.
+ */
+export async function waitForEnd(...manyPromisesToWaitFor: Promise<unknown>[]): Promise<void> {
```
EricCornelson: This and `waitForSinglePromiseToEnd` make me nervous. Do we really want to be just swallowing exceptions with no logging? I've been bitten hard by stuff like this in the past where some method was swallowing an important exception and it was near impossible to figure out why something was failing.

At the very least, I think this is a place where the name should explicitly state that we are ignoring failures. Something like `awaitIgnoreFailures()` and `awaitAllIgnoreFailures()`, and log ignored exceptions to `warn`;

digeff: You are right. That function could be easily misused.
I'll remove it so it's more difficult to ignore the failures of promises

PR #433, opened 2019-04-26
Refactor getPositionInScript and getPositionInSource to make the code and the names easier to understand

digeff:
Refactor getPositionInScript and getPositionInSource to make the code and the names easier to understand as suggested by previous feedback.

htmlToScriptPositionTranslator.ts & scriptToHtmlPositionTranslator.ts: Utilities to translate locations in script relative to the script tags (which is what source map uses), or relative to the html file with the script tags (which is what CDTP uses).

sourceMap.ts: Started using the new models to help simplify the code in sourcesMapper.ts

sourceMaps.ts, baseSourceMapTransformer.ts & eagerSourceMapTransformer.ts: Had to update to be coherent with the changes in sourceMap.ts

extensibilityPoints.ts: sourceMapTransformer is not currently being used, and the signature was failing to compile after this change, so I removed it.

### DIFF
--- a/src/chrome/internal/scripts/htmlToScriptPositionTranslator.ts
+++ b/src/chrome/internal/scripts/htmlToScriptPositionTranslator.ts
@@ -1,0 +1,20 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import { createLineNumber, createColumnNumber } from '../locations/subtypes';
+import { Position, LocationInScript } from '../locations/location';
+
+export class HtmlToScriptPositionTranslator {
+    public toPositionRelativeToScript(positionOfScriptTagInHtml: Position, positionRelativeToHtml: LocationInScript): LocationInScript {
+        // All the lines need to be adjusted by the relative position of the script in the resource (in an .html if the script starts in line 20, the first line is 20 rather than 0)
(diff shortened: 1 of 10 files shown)


src/chrome/internal/scripts/htmlToScriptPositionTranslator.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { createLineNumber, createColumnNumber } from '../locations/subtypes';
+import { Position, LocationInScript } from '../locations/location';
+
+export class HtmlToScriptPositionTranslator {
+    public toPositionRelativeToScript(positionOfScriptTagInHtml: Position, positionRelativeToHtml: LocationInScript): LocationInScript {
+        // All the lines need to be adjusted by the relative position of the script in the resource (in an .html if the script starts in line 20, the first line is 20 rather than 0)
+        const lineRelativeToScript = positionRelativeToHtml.position.lineNumber - positionOfScriptTagInHtml.lineNumber;
+
+        // The columns on the first line need to be adjusted. Columns on all other lines don't need any adjustment.
```
EricCornelson: Just curious, why is it that only the first line's columns need to be adjusted?

digeff: script.js:

> console.log('a');
> console.log('b');
> console.log('c');
> 

when turned into an inline script:

```
<script>console.log('a'); <--- all the columns in this line are shifted 8 column numbers to the right because they appear after the script tag
console.log('b'); <--- This one is not shifted
console.log('c');</script> <--- This one is not shifted either
```

PR #434, opened 2019-04-30
Fix skip files

digeff:
2 of the tests pass after thix fis.

1. We were first creating the blackbox ranges in resolveSkipFiles, and afterwards setting the new skip file status for the files, so it was ignoring the new skip file.

2. makeRegexesSkip: We were trying to create a regexp with a typescript file, instead of a javascript file

### DIFF
--- a/src/chrome/internal/features/skipFiles.ts
+++ b/src/chrome/internal/features/skipFiles.ts
@@ -104,6 +104,10 @@ export class SkipFilesLogic implements IStackTracePresentationDetailsProvider {
                 return;
             }
 
+            const newStatus = !this.shouldSkipSource(resolvedSource);
+            logger.log(`Setting the skip file status for: ${resolvedSource} to ${newStatus}`);
+            this._skipFileStatuses.setAndReplaceIfExist(resolvedSource.identifier, newStatus);
+
             for (const script of resolvedSource.scriptMapper().scripts) {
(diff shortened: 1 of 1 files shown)


roblourens:
> makeRegexesSkip: We were trying to create a regexp with a typescript file, instead of a javascript file

I can't tell from looking at the code whether this is right, but if a typescript file is marked as skipped then we should create a regex with the TS file right?

Which tests are now working?

digeff:
Two of the skip files test pass after this change (the one that uses reload fails)

src/chrome/internal/features/skipFiles.ts
```diff
@@ -104,6 +104,10 @@ export class SkipFilesLogic implements IStackTracePresentationDetailsProvider {
                 return;
             }
 
+            const newStatus = !this.shouldSkipSource(resolvedSource);
+            logger.log(`Setting the skip file status for: ${resolvedSource} to ${newStatus}`);
+            this._skipFileStatuses.setAndReplaceIfExist(resolvedSource.identifier, newStatus);
+
             for (const script of resolvedSource.scriptMapper().scripts) {
                 if (resolvedSource === script.developmentSource && script.mappedSources.length < 0) {
```
roblourens: Was this supposed to be `> 0`?

digeff: Yes it was :)

PR #436, opened 2019-05-01
Make small optimization that happens to be a workaround for the failing reload test

digeff:
Make small optimization that happens to be a workaround for the failing reload test.

Changing the position of the filter will avoid some unnecessary calls to GetPossibleBreakpoints

### DIFF
--- a/src/chrome/internal/breakpoints/features/existingBPsForJustParsedScriptSetter.ts
+++ b/src/chrome/internal/breakpoints/features/existingBPsForJustParsedScriptSetter.ts
@@ -96,9 +96,8 @@ export class ExistingBPsForJustParsedScriptSetter {
         const bpRecepieResolved = bpRecipe.resolvedWithLoadedSource(sourceWhichMayHaveBPs);
         const runtimeLocationsWhichAlreadyHaveThisBPR = debuggeeBPRecipes.map(recipe => recipe.runtimeSourceLocation);
 
-        const manyBPRecipesInScripts = await this._sourceToScriptMapper.mapBPRecipe(bpRecepieResolved);
-        const bprInScripts = manyBPRecipesInScripts.filter(b => b.location.script === justParsedScript);
-        await this.withLogging.setBPRsInScriptIfNeeded(bprInScripts, runtimeLocationsWhichAlreadyHaveThisBPR);
+        const manyBPRecipesInScripts = await this._sourceToScriptMapper.mapBPRecipe(bpRecepieResolved, script => script === justParsedScript);
+        await this.withLogging.setBPRsInScriptIfNeeded(manyBPRecipesInScripts, runtimeLocationsWhichAlreadyHaveThisBPR);
     }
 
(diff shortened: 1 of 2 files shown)


src/chrome/internal/breakpoints/features/existingBPsForJustParsedScriptSetter.ts
```diff
@@ -96,9 +96,8 @@ export class ExistingBPsForJustParsedScriptSetter {
         const bpRecepieResolved = bpRecipe.resolvedWithLoadedSource(sourceWhichMayHaveBPs);
         const runtimeLocationsWhichAlreadyHaveThisBPR = debuggeeBPRecipes.map(recipe => recipe.runtimeSourceLocation);
 
-        const manyBPRecipesInScripts = await this._sourceToScriptMapper.mapBPRecipe(bpRecepieResolved);
-        const bprInScripts = manyBPRecipesInScripts.filter(b => b.location.script === justParsedScript);
-        await this.withLogging.setBPRsInScriptIfNeeded(bprInScripts, runtimeLocationsWhichAlreadyHaveThisBPR);
+        const manyBPRecipesInScripts = await this._sourceToScriptMapper.mapBPRecipe(bpRecepieResolved, script => script === justParsedScript);
```
EricCornelson: Spelling nit: `bpRecepieResolved` should be `bpRecipeResolved`

src/chrome/internal/services/sourceToScriptMapper.ts
```diff
@@ -21,8 +22,8 @@ export class SourceToScriptMapper {
         this.doesTargetSupportColumnBreakpointsCached = this._breakpointFeaturesSupport.supportsColumnBreakpoints;
     }
 
-    public async mapBPRecipe(bpRecipe: BPRecipeInLoadedSource<ConditionalPause | AlwaysPause>): Promise<BPRecipeInScript[]> {
-        const tokensInManyScripts = bpRecipe.location.tokensWhenMappedToScript();
+    public async mapBPRecipe(bpRecipe: BPRecipeInLoadedSource<ConditionalPause | AlwaysPause>, onlyKeepIfScript: (script: IScript) => boolean = () => true): Promise<BPRecipeInScript[]> {
+        const tokensInManyScripts = bpRecipe.location.tokensWhenMappedToScript().filter(tokens => onlyKeepIfScript(tokens.script));
```
EricCornelson: Just to clarify on what this filter is doing. It looks like this is accounting for the case where we have a resource that could map to multiple script sources, and we only care about the script that was recently parsed, correct? And the only way, then, to check if the breakpoint is in the script we just parsed is to resolve all of them associated with that resource?

Do we have to do the mapping resolve every time we process any breakpoint associated with that resource?

src/chrome/internal/breakpoints/features/existingBPsForJustParsedScriptSetter.ts
digeff: Fixed

PR #438, opened 2019-05-03
Fixes associated with chrome launch tests

EricCornelson:
Added code to handle attach requests in ConnectingCDA, and updated vscode-nls-dev to fix windows path separator bug (this was breaking localization and causing the "messages not found" error).

### DIFF
--- a/src/chrome/client/chromeDebugAdapter/connectingCDA.ts
+++ b/src/chrome/client/chromeDebugAdapter/connectingCDA.ts
@@ -10,6 +10,9 @@ import { ChromeConnection } from '../../chromeConnection';
 import { ConnectedCDA, ConnectedCDAProvider } from './connectedCDA';
 import { ConnectedCDAConfiguration } from './cdaConfiguration';
 import { ITelemetryPropertyCollector } from '../../../telemetry';
+import { ScenarioType } from './unconnectedCDA';
+import { IAttachRequestArgs } from '../../../debugAdapterInterfaces';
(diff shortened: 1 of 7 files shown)


src/chrome/client/chromeDebugAdapter/connectingCDA.ts
```diff
@@ -25,8 +28,18 @@ export class ConnectingCDA extends BaseCDAState {
...
+        if (this._configuration.scenarioType === ScenarioType.Launch) {
+            logger.verbose('[ConnectingCDA]: Launching debugee...');
+            const result = await this._debuggeeLauncher.launch(this._configuration.args, telemetryPropertyCollector);
+            await this._chromeConnection.attach(result.address, result.port, result.url, this._configuration.args.timeout, this._configuration.args.extraCRDPChannelPort);
+        }
+        else if (this._configuration.scenarioType === ScenarioType.Attach) {
+            logger.verbose('[ConnectingCDA]: Attaching to an existing instance of debugee...');
+            const attachArgs = <IAttachRequestArgs>this._configuration.args;
+            await this._chromeConnection.attach(attachArgs.address, attachArgs.port, attachArgs.url, attachArgs.timeout, attachArgs.extraCRDPChannelPort);
+        }
```
digeff: I'd add an else case and throw an exception or something if it's neither Launch or Attach (which should be impossible, yet I've seen impossible things happening in software before...)

PR #439, opened 2019-05-06
Add new typed operators to be used instead of boolean operators when enabling tslint rule

digeff:
Add new typed operators to be used instead of boolean operators when enabling tslint rule

To see how these operators are used check this change: https://github.com/Microsoft/vscode-chrome-debug-core/compare/v2...digeff:extra_tslint_rules?expand=1

We are going to enable the strict-boolean-expressions rule which doesn't allow to use boolean operators on things that are not strictly boolean. We are going to use these operators instead.

The types are choosen to minimize the chance of introducing bugs (even though calling isDefined on a string makes sense, if we were using if (string) before, the equivalent is to call isNotEmpty, so the type of isDefined doesn't accept strings to force you to call isNotEmpty instead). This makes sense while we enable the rule, to make sure that we preserve backwards-compatibility.

In the future it'll make sense to call isDefined for a string, so we'll probably add the type to allow that after all the PRs of this tslint rule have been merged.

### DIFF
--- a/src/chrome/utils/typedOperators.ts
+++ b/src/chrome/utils/typedOperators.ts
@@ -1,0 +1,100 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+/**
+ * These file includes functions normally implemented by JavaScript operators. The reason we use these functions
+ * is to have better control over the typing of the functions, and use that to help reduce the chance of introducing errors
+ * e.g.: We on purpose do not allowe isDefined() to be called for strings, because there is a fair chance that we meant to use isNotEmpty instead
+ */
+
+/**
+ * Returns whether the object or number is an actual object or number (and is not undefined)
+ */
+export function isDefined<T extends object | number>(objectOrUndefined: T | undefined): objectOrUndefined is T {
+    return objectOrUndefined !== undefined;
+}
+
+/**
+ * Returns whether the object or number is an actually undefined
+ */
+export function isUndefined<T extends object | number>(objectOrUndefined: T | undefined): objectOrUndefined is undefined {
+    return objectOrUndefined === undefined;
+}
+
+/**
+ * Returns whether the nullable object or string is not null
+ */
+export function isNotNull<T extends object | string>(objectOrNull: T | null): objectOrNull is T {
+    return objectOrNull !== null;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


src/chrome/utils/typedOperators.ts
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * These file includes functions normally implemented by JavaScript operators. The reason we use these functions
+ * is to have better control over the typing of the functions, and use that to help reduce the chance of introducing errors
+ * e.g.: We on purpose do not allowe isDefined() to be called for strings, because there is a fair chance that we meant to use isNotEmpty instead
+ */
+
+/**
+ * Returns whether the object or number is an actual object or number (and is not undefined)
+ */
+export function isDefined<T extends object | number>(objectOrUndefined: T | undefined): objectOrUndefined is T {
```
roblourens: Why `object | number`? Does that include string? Do you really need the type constraint?

src/chrome/utils/typedOperators.ts
```diff
@@ -0,0 +1,100 @@
+/*---------------------------------------------------------
```
roblourens: Does this file get excluded from the lint rule? (just wondering how this works)

src/chrome/utils/typedOperators.ts
roblourens: Oh I read the comment, got it

src/chrome/utils/typedOperators.ts
digeff: It doesn't.
Instead of doing: !!stringOrUndefined we do: stringOrUndefined === undefined 
Instead of doing if (objectOrNull) we do: if (objectOrNull !== null)

All the code that appears on this file could appear directly on the code. I found it easier to replace
if (obj) for if (isNotEmpty(obj)) rather than if (obj !== undefined) because depending on the type if (obj) needs to be replaced by something different:
if obj is boolean? then you should do if (obj === true)
if obj is string?  then you should do if (obj !== undefined && obj !== '')
if obj is an object then you should do if (obj !== undefined)

Because the functions we are creating have specific types, if I try to do the wrong replacement, it won't compile. If I do it directly on the files, I won't get any warnings.

In some cases like hasMatches(match) I also find that more readable than match !== null

roblourens: Ah ok got it

PR #440, opened 2019-05-06
TSLint - strict boolean expressions - 1

digeff:


src/chrome/cdtpDebuggee/protocolParsers/cdtpLocationParser.ts
```diff
@@ -6,6 +6,7 @@ import { Position, LocationInScript } from '../../internal/locations/location';
 import { createColumnNumber, createLineNumber } from '../../internal/locations/subtypes';
 import { CDTPScriptsRegistry } from '../registries/cdtpScriptsRegistry';
 import { Protocol as CDTP } from 'devtools-protocol';
+import _ = require('lodash');
```
roblourens: Why `require`?

digeff: Mistake :) Changing...

roblourens:
Conflicts from the other PR now

PR #441, opened 2019-05-06
TSLint - strict boolean expressions - 2

digeff:


src/chrome/chromeDebugSession.ts
```diff
@@ -15,6 +15,8 @@ import { ExecutionTimingsReporter, StepProgressEventsEmitter, IObservableEvents,
 import { ChromeDebugAdapter } from './client/chromeDebugAdapter/chromeDebugAdapterV2';
 import { CommandText } from './client/requests';
 import { IExtensibilityPoints } from './extensibility/extensibilityPoints';
+import { isNotEmpty, isTrue, isDefined, ifDefinedDo } from './utils/typedOperators';
+import _ = require('lodash');
```
roblourens: More `require`

digeff: Fixed

PR #442, opened 2019-05-06
TSLint - strict boolean expressions - 3

digeff:


src/sourceMaps/sourceMap.ts
```diff
@@ -53,20 +55,27 @@ class SourcePathMappingCalculator {
         // is not always needed.
         return this._normalizedSourcesInOrder.map((inferredPath: IResourceIdentifier, i: number) => {
             const originalSource = this._originalSourcesInOrder[i];
-            const originalPath = this._originalSourceRoot
+            const originalPath = isNotEmpty(this._originalSourceRoot)
                 ? sourceMapUtils.getFullSourceEntry(this._originalSourceRoot, originalSource)
                 : originalSource;
+
+            let starPosition;
```
EricCornelson: `startPosition`?

digeff: Fixed

PR #443, opened 2019-05-06
TSLint - strict boolean expressions - 4

digeff:

### DIFF
--- a/src/chrome/consoleHelper.ts
+++ b/src/chrome/consoleHelper.ts
@@ -8,9 +8,8 @@ import * as variables from './variables';
 import { CodeFlowStackTrace } from './internal/stackTraces/codeFlowStackTrace';
 import { IExceptionDetails } from './cdtpDebuggee/eventsProviders/cdtpExceptionThrownEventsProvider';
 import { functionDescription } from './internal/stackTraces/callFramePresentation';
-import isUndefined = require('lodash/isUndefined');
 import * as _ from 'lodash';
-import { isNotEmpty, isFalse, isDefined } from './utils/typedOperators';
+import { isNotEmpty, isFalse, isDefined, isUndefined } from './utils/typedOperators';
 
 export function formatExceptionDetails(e: IExceptionDetails): string {
     if (isUndefined(e.exception)) {
--- a/src/chrome/consoleHelper.ts
+++ b/src/chrome/consoleHelper.ts
@@ -8,14 +8,18 @@ import * as variables from './variables';
 import { CodeFlowStackTrace } from './internal/stackTraces/codeFlowStackTrace';
 import { IExceptionDetails } from './cdtpDebuggee/eventsProviders/cdtpExceptionThrownEventsProvider';
 import { functionDescription } from './internal/stackTraces/callFramePresentation';
+import isUndefined = require('lodash/isUndefined');
+import _ = require('lodash');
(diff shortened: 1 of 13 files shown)


src/chrome/consoleHelper.ts
```diff
@@ -110,20 +113,20 @@ function resolveParams(args: CDTP.Runtime.RemoteObject[], skipFormatSpecifiers?:
...
-                currentCollapsedStringArg += (currentCollapsedStringArg ? ' ' + formatted : formatted);
+                currentCollapsedStringArg += (currentCollapsedStringArg === '' ? ' ' + formatted : formatted);
             }
-        } else if (formatSpec) {
+        } else if (isNotEmpty(formatSpec)) {
             // `formatted` is an object - split currentCollapsedStringArg around the current formatSpec and add the object
             const curSpecIdx = currentCollapsedStringArg.indexOf('%' + formatSpec);
             const processedPart = currentCollapsedStringArg.slice(0, curSpecIdx);
-            if (processedPart) {
+            if (processedPart !== '') {
```
mrcrane: Use isEmpty?

src/utils.ts
```diff
@@ -193,31 +196,27 @@ export function forceForwardSlashes(aUrl: string): string {
...
-
-    if (aPath.match(/file:\/\/\/[A-Za-z]:/)) {
+    if (hasMatches(aPath.match(/file:\/\/\/[A-Za-z]:/))) {
         const prefixLen = 'file:///'.length;
         aPath =
             'file:///' +
             aPath[prefixLen].toLowerCase() +
             aPath.substr(prefixLen + 1);
-    } else if (aPath.match(/^[A-Za-z]:/)) {
+    } else if (aPath.match(/^[A-Za-z]:/) !== null) {
```
mrcrane: Use isNull?

src/utils.ts
```diff
@@ -240,11 +239,11 @@ export function stripTrailingSlash(aPath: string): string {
  * when passing on a failure from a Promise error handler.
  * @param msg - Should be either a string or an Error
  */
-export function errP(msg: string | Error): Promise<never> {
+export function errP(msg?: string | Error): Promise<never> {
     const isErrorLike = (thing: any): thing is Error => !!thing.message;
 
     let e: Error;
-    if (!msg) {
+    if (msg === undefined) {
```
mrcrane: use isUndefined?

src/utils.ts
```diff
@@ -489,7 +457,7 @@ export function pathToRegex(aPath: string, guid = ''): string {
         aPath = escapeRegexSpecialChars(fileUrlPrefix) + aPath;
     }
 
-    if (guid) {
+    if (guid !== '') {
```
mrcrane: use isEmpty?

src/chrome/consoleHelper.ts
digeff: Changed to isNotEmpty

src/utils.ts
digeff: Changed to hasMatches

src/utils.ts
digeff: Changed to isNotEmpty

src/utils.ts
digeff: Changed method to only accept strings, and changed to isEmpty

PR #444, opened 2019-05-06
TSLint - strict boolean expressions - 5

digeff:


roblourens:
Build error

PR #446, opened 2019-05-22
Fix stop bug, and logging bug

digeff:
- We were calling this._debuggeeLauncher.stop two times, once inside TerminatingCDA.disconnect and one inside ChromeConnection.close. We remove the first call.
At the moment it also make sense to me to first terminate the session before calling the shutdown method, so I changed that order too. Similar with the TerminatedEvent

- The logging bug in src/chrome/internal/services/logging.ts. Variable logFilePath  was a boolean when it should've been a string


EricCornelson:
Looks good

PR #448, opened 2019-05-28
Made logging more granular

digeff:
The previous logging logic was logging only the start of the methods, and the end of the methods. It was logging the start in not the exact order where it was executed (because we waited to know if the method was sync or not).

This small changes logs 2 steps for each sync function, and 3 steps for each async function which should match 100% the exact order in which things are executed.
sync: Start and End
async: Start, Method returns the promise synchronically, Async End

This granularity was used to investigate some bugs in our integration tests

### DIFF
--- a/src/chrome/logging/methodsCalledLogger.ts
+++ b/src/chrome/logging/methodsCalledLogger.ts
@@ -60,6 +60,7 @@ export class MethodsCalledLogger<T extends object> {
         const handler = {
             get: <K extends keyof T>(target: T, propertyKey: K, receiver: any) => {
                 const originalPropertyValue = target[propertyKey];
+                // tslint:disable-next-line: strict-type-predicates
                 if (typeof originalPropertyValue === 'function') {
                     return (...args: any) => {
                         const callId = this.generateCallId();
--- a/src/chrome/logging/methodsCalledLogger.ts
+++ b/src/chrome/logging/methodsCalledLogger.ts
@@ -24,18 +24,21 @@ export class ReplacementInstruction {
 export interface IMethodsCalledLoggerConfiguration {
     readonly replacements: ReplacementInstruction[];
 
-    decideWhetherToWrapMethodResult(methodName: string | symbol | number, args: unknown[], result: unknown, wrapWithName: (name: string) => void): void;
-    decideWhetherToWrapEventEmitterListener(receiverName: string, methodName: string | symbol | number, args: unknown[], wrapWithName: (name: string) => void): void;
+    customizeResult(methodName: string | symbol | number, args: unknown[], result: unknown): unknown;
+    customizeArgumentsBeforeCall(receiverName: string, methodName: string | symbol | number, args: unknown[]): void;
 }
 
 export class MethodsCalledLoggerConfiguration implements IMethodsCalledLoggerConfiguration {
 
     public constructor(public readonly containerName: string, private _replacements: ReplacementInstruction[]) { }
 
-    public decideWhetherToWrapMethodResult(_methodName: string | symbol | number, _args: unknown[], _result: unknown, _wrapWithName: (name: string) => void): void { }
-    public decideWhetherToWrapEventEmitterListener(receiverName: string, methodName: string | symbol | number, args: unknown[], wrapWithName: (name: string) => void): void {
-        if (methodName === 'on') {
-            wrapWithName(`(${receiverName} emits ${args[0]})`);
+    public customizeResult(_methodName: string | symbol | number, _args: unknown[], result: unknown): unknown {
+        return result;
+    }
+
+    public customizeArgumentsBeforeCall(receiverName: string, methodName: string | symbol | number, args: object[]): void {
+        if (methodName === 'on' && args.length >= 2) {
+            args[1] = new MethodsCalledLogger(this, args[1], `(${receiverName} emits ${args[0]})`).wrapped();
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/chrome/logging/methodsCalledLogger.ts
```diff
@@ -59,38 +58,48 @@ export class MethodsCalledLogger<T extends object> {
...
                         try {
                             if (propertyKey === 'on' && args.length >= 2) {
                                 let listenerPossiblyWrapped = args[1];
                                 this._configuration.decideWhetherToWrapEventEmitterListener(this._objectToWrapName, propertyKey, args, name => listenerPossiblyWrapped = new MethodsCalledLogger(this._configuration, args[1], name).wrapped());
                                 args[1] = listenerPossiblyWrapped;
                             }
 
+                            this.logCallStart(propertyKey, args, callId);
                             const result = originalPropertyValue.apply(target, args);
                             if (!result || !result.then) {
```
EricCornelson: Does this pass the boolean expressions lint rule?

src/chrome/logging/methodsCalledLogger.ts
```diff
@@ -59,38 +58,48 @@ export class MethodsCalledLogger<T extends object> {
...
                                 return result.then((promiseResult: unknown) => {
                                     this.logCall(propertyKey, Synchronicity.Async, args, Outcome.Succesful, promiseResult, callId);
-                                    let resultPossiblyWrapped = promiseResult;
-                                    this._configuration.decideWhetherToWrapMethodResult(propertyKey, args, promiseResult, name => resultPossiblyWrapped = new MethodsCalledLogger(this._configuration, <object>promiseResult, name).wrapped());
-                                    return resultPossiblyWrapped;
+                                    if (promiseResult === target) {
+                                        return receiver;
+                                    } else {
+                                        let resultPossiblyWrapped = promiseResult;
+                                        this._configuration.decideWhetherToWrapMethodResult(propertyKey, args, promiseResult, name => resultPossiblyWrapped = new MethodsCalledLogger(this._configuration, <object>promiseResult, name).wrapped());
```
EricCornelson: Two notes on the wrapping logic here.

1. This seems like a really complicated way to wrap the name for emitters (create a variable, pass it through a closure, then reassign/return it), and we repeat this logic several times. Could we just have the method return a string and assign it directly rather than passing the variable through a closure?

2. Is there really a need to make this logic so abstract/configurable? I only see one implementation of `IMethodsCalledLoggerConfiguration` and the method `decideWhetherToWrapEventEmitterListener` seems extremely specific to this one use case. Is there ever a case where we'd override this behavior, or really need another implementation of that interface?

src/chrome/logging/methodsCalledLogger.ts
digeff: What do you mean?

src/chrome/logging/methodsCalledLogger.ts
digeff: 1. Do you mean returning the name for the object instead of assigning the full MethodsCalledLogger?
2. In the test we manually wrap the Wizard classes in the MethodsLogger to log the actions. A better approach than doing that manually, could be to automatically detect when a Wizard is returned, and wrap it with a proxy automatically. If we do that, we'd use this method/interface

src/chrome/logging/methodsCalledLogger.ts
EricCornelson: Didn't we just enable the `strict-boolean-expressions` lint rule that ended up requiring lots of `isDefined` and similar calls? But here we're just testing `result` the normal "javascripty" way. Not a problem, just curious.

src/chrome/logging/methodsCalledLogger.ts
EricCornelson: In response to (1) Why not just let it return the object, wrapped or not, with any name changes the implementation feels like adding? `decideWhetherToWrapMethodResult` is already taking the `result` object as a param, why can't he just return the result unaltered if he chose not to wrap it, and return the wrapped object (the passed wrapping function would return the wrapped object rather than void). In addition to that, it looks like the wrapping function is mostly the same in all 3 places. We already have `wrapWithMethodLogger` which looks like it could do what we need for that. In which case, if we're passing (essentially) the same function for every invocation, why pass it? Just use `wrapWithMethodLogger`, or maybe pass it in the constructor to `MethodsCalledLoggerConfiguration` if we think that would ever change.

It all seems like a bit of a moot point anyway, since the only (current) implementation of `decideWhetherToWrapMethodResult` is a no-op, and I tried setting a breakpoint on line 69 and did just about everything I could in the debug adapter and couldn't get inside that block.

I'll approve and let you decide what you want to do with it, but It just seems like a lot of really confusing code that isn't actually doing anything \*right now\*, for the promise of _maybe_ being able to use it for something pretty minor in the future, and we've already spent way too much time on it. YAGNI.

src/chrome/logging/methodsCalledLogger.ts
digeff: You are right, I missed that.
result is of type any, so it doesn't trigger that rule, so tslint is not complaining...

src/chrome/logging/methodsCalledLogger.ts
digeff: I think "return the object, wrapped or not" is a great idea! I'll change the code...

P.S.: decideWhetherToWrapMethodResult is overwritten/used in vscode-chrome-debug\test\int\utils\logging.ts

PR #452, opened 2019-06-07
Fixes for DTL tests

digeff:
- package.json: Chaged to version 10 so it's easier to determine from logs and telemetry whether this is v2 or not...

- pauseActionsPriorities.ts: All actions should be in this list. PausedBecauseAsyncCallWasScheduled was missing it, so we added it.

- syncStepping.ts: Things were breaking when we hit a breakpoint while stepping. The SyncStepping class ended up stuck on the CurrentlyStepping state and it didn't allow further stepping. We also improved the toString of the stepping states and added some more logging

- evaluateRequestHandler.ts: We were missing the frame parameter. Because the parameter is optional, the compiler didn't complain.

- printing.ts: Improved the toString of some classes to add more information to the logs

- fallbackToClientPathTransformer.ts: response.body.filePath is of type string yet we were using it as it were a IResourceIdentifier. response.body.filePath has type any so the compiler didn't complain when we did this.

- tsconfig.json: Changed to es2017 which is fully supported by node 9.11.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "vscode-chrome-debug-core",
   "displayName": "vscode-chrome-debug-core",
-  "version": "6.7.35",
+  "version": "10.7.35",
   "description": "A library for building VS Code debug adapters for targets that support the Chrome Remote Debug Protocol",
   "repository": {
     "type": "git",
--- a/src/chrome/utils/printing.ts
+++ b/src/chrome/utils/printing.ts
@@ -4,18 +4,10 @@ function _printClassDescription(this: Function): string {
     return isNotEmpty(this.name) ? `class ${this.name}` : Function.toString.call(this);
 }
 
-function _printInstanceDescription(this: object): string {
-    return isNotEmpty(this.constructor.name) ? `${this.constructor.name}` : 'an anonymous object';
+export function printInstanceDescription(this: object): string {
+    return `${this.constructor.name}`;
 }
 
 export function printClassDescription(functionConstructor: Function) {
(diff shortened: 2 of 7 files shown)


src/chrome/utils/printing.ts
```diff
@@ -4,6 +4,18 @@ function _printClassDescription(this: Function): string {
...
+}
+
 export function printClassDescription(functionConstructor: Function) {
     functionConstructor.toString = _printClassDescription;
+
+    /*
+     * If the class has the default toString method with returns [object Object] change it for our custom method
+     * that returns the class name instead
+     */
+    if (functionConstructor.prototype.toString === Object.prototype.toString) {
```
roblourens: How about just having these class extend a superclass, or implement toString to call a function? This seems like an unusual pattern that could confuse somebody.

EricCornelson: Agree. Especially because it looks like this will overwrite any existing toString method on the class, which would definitely be surprising behavior. A simple function isn't any more code than a decorator in this case, but much easier to read and understand.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "vscode-chrome-debug-core",
   "displayName": "vscode-chrome-debug-core",
-  "version": "6.7.35",
+  "version": "10.7.35",
```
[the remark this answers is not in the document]
digeff: I'm not planning to release this number on production. This is just to clearly see it while we are developing it...

src/chrome/utils/printing.ts
digeff: Changed to a method on the class...

PR #454, opened 2019-06-12
Fix close chrome issue

digeff:
When closing chrome we were exiting the debug adapter before it had a chance to properly clean-up everything.

Now when we close chrome we only do the terminate processing, and we wait for the client to send us the disconnect request before actually trying to disconnect/exit/kill the debug adapter


src/chrome/client/chromeDebugAdapter/connectedCDA.ts
```diff
@@ -28,21 +29,22 @@ export class ConnectedCDA extends BaseCDAState {
...
         super(requestHandlerDeclarers, {
             'initialize': () => { throw new Error('The debug adapter is already initialized. Calling initialize again is not supported.'); },
             'launch': () => { throw new Error("Can't launch  to a new target while connected to a previous target"); },
             'attach': () => { throw new Error("Can't attach to a new target while connected to a previous target"); },
-            'disconnect': async () => {
+            'disconnect': async (_args: DebugProtocol.DisconnectArguments) => {
+                // TODO: Add support for args.terminateDebuggee
                 this._ignoreNextDisconnectedFromWebSocket = true;
-                await this.disconnect(TerminatingReason.DisconnectedFromWebsocket);
+                await this.terminate(TerminatingReason.DisconnectedFromWebsocket);
```
roblourens: I don't think you should handle 'disconnect' as 'terminate(' because the DAP has those as different requests.

src/chrome/client/chromeDebugAdapter/baseCDAState.ts
```diff
@@ -20,7 +24,19 @@ export abstract class BaseCDAState implements IDebugAdapterState {
     }
 
     public async processRequest(requestName: CommandText, args: unknown): Promise<unknown> {
-        return await this._requestProcessor.processRequest(requestName, args);
+        try {
+            return await this._requestProcessor.processRequest(requestName, args);
+        } finally {
+            if (requestName === 'disconnect') {
+                // Disconnect shuts down the session forcefully after executing the available request handler
+                await this.shutdown();
```
EricCornelson: Will this kill the process before we have a chance to send the response to the 'disconnect' request?

digeff: At the moment it has 100ms to send the response: https://github.com/microsoft/vscode-chrome-debug-core/pull/454

roblourens: It seems weird for baseCDAstate to call shutdown like this, why not call it from the actual handler?

digeff: Based on the Debug session end description here: https://microsoft.github.io/debug-adapter-protocol/overview

"The disconnect request is expected to terminate the debuggee (and any child processes) forcefully"

Makes me think that the disconnect could potentially be sent in any state, to kill the debug adapter. This was my way of making sure that all states know how to disconnect

roblourens: Oh ok. disconnect is probably special in that regard then.

src/chrome/client/chromeDebugAdapter/baseCDAState.ts
```diff
@@ -4,13 +4,20 @@ import { CommandText } from '../requests';
...
 export abstract class BaseCDAState implements IDebugAdapterState {
     private readonly _requestProcessor: RequestProcessor;
+    protected abstract readonly _session: ISession;
 
-    constructor(requestHandlerDeclarers: ICommandHandlerDeclarer[], requestHandlerMappings: RequestHandlerMappings) {
-        const allDeclarers = requestHandlerDeclarers.concat({ getCommandHandlerDeclarations: () => CommandHandlerDeclaration.fromLiteralObject(requestHandlerMappings) });
+    constructor(
+        requestHandlerDeclarers: ICommandHandlerDeclarer[],
+        requestHandlerMappings: RequestHandlerMappings) {
+        const requestHandlerMappingsWithDefault = { disconnect: () => this.shutdown(), ...requestHandlerMappings };
```
EricCornelson: I'd add a comment here explaining what you're doing (adding a default disconnect handler). It took me a while to parse out what was actually happening in this code, and at first it looked to me like this would just overwrite whichever 'disconnect' handler got passed.

digeff: Added comment

src/chrome/client/chromeDebugAdapter/connectedCDA.ts
digeff: Added disconnect method

PR #458, opened 2019-06-19
More fixes for node v2

digeff:
- Renamed setAndReplaceIfExist to setAndReplaceIfExists

- terminatingCDA.ts: The terminatedCDA wasn't installed, so it failed when it received any requests from the client
- frameParser.ts: Created utility because we consume frames via id both in completions and in evaluate
- completionsRequestHandler.ts: We hadn't implemented support for the completions request
- resourceIdentifier.ts: Setting a breakpoint on c:\myProj\my file name.js was failing because we weren't escaping the spaces when converting it to a file url
- sourceMapFactory.ts: err is either null or an error, instead of undefined
- sourceMaps.ts: We weren't initializing this._authoredPathToSourceMap. This is needed when we use a typescript.ts file as a startup file in node, so we can convert it into a javascript file
- baseSourceMapTransformer.ts: We need to use this class before we are connected to the client, to convert a typescript.ts startup file to a javascript file. Modified the constructor parameter so we can create a "fake" version of this class before we are connected to the CRDP websocket

### DIFF
--- a/src/chrome/internal/completions/completionsRequestHandler.ts
+++ b/src/chrome/internal/completions/completionsRequestHandler.ts
@@ -1,0 +1,24 @@
+import { ChromeDebugLogic } from '../../chromeDebugAdapter';
+import { ICommandHandlerDeclaration, CommandHandlerDeclaration, ICommandHandlerDeclarer } from '../features/components';
+import { injectable, inject } from 'inversify';
+import { DebugProtocol } from 'vscode-debugprotocol';
+import { TYPES } from '../../dependencyInjection.ts/types';
+import { FrameParser } from '../../client/frameParser';
+
+@injectable()
+export class CompletionsRequestHandler implements ICommandHandlerDeclarer {
+    public constructor(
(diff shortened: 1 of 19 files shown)


src/chrome/internal/completions/completionsRequestHandler.ts
```diff
@@ -0,0 +1,24 @@
...
+import { injectable, inject } from 'inversify';
+import { DebugProtocol } from 'vscode-debugprotocol';
+import { TYPES } from '../../dependencyInjection.ts/types';
+import { FrameParser } from '../../client/frameParser';
+
+@injectable()
+export class CompletionsRequestHandler implements ICommandHandlerDeclarer {
+    public constructor(
+        @inject(TYPES.ChromeDebugLogic) protected readonly _chromeDebugAdapter: ChromeDebugLogic,
+        private readonly _frameParser: FrameParser) { }
```
roblourens: Should this have an @inject tag?

digeff: Because we are using the experimental metadata, I think that you don't need the inject for classes (only for interfaces)

PR #459, opened 2019-06-24
Fixes to make the node v2 tests pass

digeff:
**List of issues and fixes in the PR, and in which files it happened:**
1. We were accidentally creating two instances of _scriptsRegistry: CDTPScriptsRegistry because some of the uses didn't had the @inject
    - src/chrome/cdtpDebuggee/eventsProviders/cdtpConsoleEventsProvider.ts
    - src/chrome/cdtpDebuggee/features/cdtpScriptSourcesRetriever.ts
    - src/chrome/chromeDebugAdapter.ts
    - src/chrome/internal/exceptions/pauseOnException.ts
    - src/transformers/eagerSourceMapTransformer.ts

2. We were setting breakpoints for .js files twice because we considered the runtimeSource and developmentSource to be different scripts. Now we make sure that if they are the same file, we share the same instance to avoid this duplication
    - src/chrome/cdtpDebuggee/eventsProviders/cdtpOnScriptParsedEventProvider.ts
    - src/chrome/internal/sources/loadedSourceToScriptRelationship.ts: Unified RuntimeSourceOf and DevelopmentSourceOf to have a shared instance, and eliminate the duplicated breakpoints we were adding
    - src/chrome/internal/sources/unidentifiedLoadedSource.ts: Return a single relationship for an unmapped file

3. this._exception was never set, so the logic was not showing the exception scope. Updated code to work on v2
    - src/chrome/chromeDebugAdapter.ts
    - src/chrome/internal/stackTraces/currentStackTraceProvider.ts: Add methods to access the exception on the last paused event

4. _.defaultsTo(value, default): uses the original value if value === "". We needed to use the default in that case
    - src/chrome/chromeUtils.ts
    - src/chrome/consoleHelper.ts
    - src/chrome/utils/typedOperators.ts
    - src/sourceMaps/sourceMapUtils.ts

5. Node debug needs to customize how the internal node sources are printted, so we created the ISourceToClientConverter interface and TYPES.SourceToClientConverter so it's possible to override that functionality
    - src/chrome/client/eventsToClientReporter.ts
    - src/chrome/client/locationInSourceToClientConverter.ts
    - src/chrome/client/sourceToClientConverter.ts
    - src/chrome/dependencyInjection.ts/bind.ts
    - src/chrome/dependencyInjection.ts/types.ts
    - src/chrome/internal/breakpoints/features/bpRecipieStatusToClientConverter.ts
    - src/chrome/internal/breakpoints/features/setBreakpointsRequestHandler.ts
    - src/chrome/internal/sources/sourceRequestHandler.ts
    - src/chrome/internal/stackTraces/stackTraceRequestHandler.ts
    - src/index.ts: Export classes and interfaces that we need to reference from node debug

6. Implemented logic to set an heuristic breakpoints when we set a breakpoint on a source for which we don't have any loaded script that is related:
    - src/chrome/internal/breakpoints/baseMappedBPRecipe.ts: Extracted mappedToUrlRegexp to it's own function, so we can use it for the Breakpoints heuristic logic
    - src/chrome/internal/breakpoints/diBindings.ts
    - src/chrome/internal/breakpoints/diTypes.ts
    - src/chrome/internal/breakpoints/features/bpAtNotLoadedScriptViaHeuristicSetter.ts: Logic to set the heuristic breakpoint
    - src/chrome/internal/breakpoints/features/existingBPsForJustParsedScriptSetter.ts: Remove breakpoint heuristics after the breakpoint is set properly
    - src/chrome/internal/breakpoints/features/hitCountBreakpointsSetter.ts
    - src/chrome/internal/breakpoints/features/onPausedForBreakpointCallback.ts
    - src/chrome/internal/breakpoints/features/singleBreakpointSetter.ts: Add and remove breakpoint heuristics when neccesary
    - src/chrome/internal/breakpoints/features/singleBreakpointSetterWithHitCountSupport.ts
    - src/chrome/internal/locations/location.ts: We modified the ILocation interface to be able to use a IHasSourceMappingInformation in places where we used to use a Script
    - src/chrome/internal/locations/mappedTokensInScript.ts: A lot of the source mapping logic expects a IScript. We don't have one for when we set an heuristics. We modified the logic to work both with an IScript and a IHasSourceMappingInformation so we can make use this code for breakpoint heuristics too
    - src/chrome/internal/locations/rangeInScript.ts
    - src/chrome/internal/scripts/IHasSourceMappingInformation.ts: Interface to use instead of a script when we are setting an heuristic breakpoint
    - src/chrome/internal/scripts/script.ts: Refactor the code so IScript is compatible with IHasSourceMappingInformation
    - src/chrome/internal/scripts/sourcesMapper.ts: Refactor the code so we can use a IHasSourceMappingInformation instead of an IScript when sourcemapping
    - src/chrome/internal/sources/identifiedLoadedSource.ts
    - src/chrome/internal/sources/loadedSource.ts
    - src/sourceMaps/sourceMap.ts: Refactor to support the IHasSourceMappingInformation interface
    - src/sourceMaps/sourceMaps.ts: The heuristic logic relies on reading the source map files that we find by looking thorugh the outFiles launch.json parameter. We copied some logic from v1 that we use after the EagerSourceMapTransformer pre-loads all the source-maps, to find them while setting the heuristic breakpoints, and use the mapping if appropiated.

7. Not all generated javascript files have a SourceMapURL property. Updated logic to be able to use source maps, as long as all the required information is on the source maps
    - src/transformers/baseSourceMapTransformer.ts: If a SourceMapURL is missing, try to find the information in the pre-loaded source maps

8. Debugger was getting stuck when we didn't recognize a breakpoint, it didn't know whether to resume or pause. The pause logic ignores all actions that are instances of NoActionIsNeededForThisPause, so making NoRecognizedBreakpoints subclass it fixed the issue.
    - src/chrome/internal/breakpoints/features/bpRecipeAtLoadedSourceLogic.ts

9. Breakpoints wheren't hitting when the source-map had absolute paths, because the logic was concatenating them with the computed root
    - src/sourceMaps/sourceMap.ts: Now we handle absolute paths correctly


src/chrome/internal/breakpoints/features/bpAtNotLoadedScriptViaHeuristicSetter.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    public toString(): string {
+        return `Source with map: ${this._sourceMap.generatedPath}`;
+    }
+}
+
+/**
+ * We need a runtime location to create a BPRecipeInRegexp. We don't have one, so we pass a this instance instead.
+ */
+class NoLoadedSourceAvailable implements ILoadedSource<CDTPScriptUrl> {
```
EricCornelson: Do we really need to create a dummy instance of this interface just to satisfy an arbitrary contract elsewhere in the code?  Would it be simpler to create an instance of `Location` that didn't care about whether we have a loaded source or not?

src/chrome/internal/breakpoints/features/bpAtNotLoadedScriptViaHeuristicSetter.ts
```diff
@@ -0,0 +1,130 @@
...
+        return new LocationInUrl(<IResourceIdentifier<CDTPScriptUrl>>sourceIdentifier, requestedBP.location.position);
+    }
+}
+
+/**
+ * We use this file when we have a typescript file, for which we found a source-map using the EagerSourceMapTransformer before the correspondant script
+ * was loaded by CDTP, so we need to use the source-mapping logic, yet we don't have an instance of IScript to do it. We use this instance instead,
+ * that will correctly help map typescript files to javascript files for non .html files.
+ */
+export class SourceWithSourceMap implements IHasSourceMappingInformation {
```
EricCornelson: Similar thing with this class. Is it really necessary to create a spoof instance of this class just to tell us to start at 0,0?

src/chrome/internal/breakpoints/features/bpAtNotLoadedScriptViaHeuristicSetter.ts
digeff: Currently the type of the parameter we need to send is LocationInLoadedSource, so we'd need to pass something that satisfies that type...

src/chrome/internal/breakpoints/features/bpAtNotLoadedScriptViaHeuristicSetter.ts
digeff: How would you do this differently?

PR #466, opened 2019-07-01
Fix promises

digeff:
this._promiseRejectionsStrategy.shouldPauseOnRejections was false in chrome-debug so we weren't showing the error message when pausing due to a promise, because we used the default logic of pausing with a debugger paused statement.
Change the logic so if this._promiseRejectionsStrategy.shouldPauseOnRejections is false, we are going to try to auto-resume the promise rejection (for node-debug).
For chrome-debug this._promiseRejectionsStrategy.shouldPauseOnRejections should always be true (to match v1 behavior), so commented the code for the time being, so we always pause on rejections

cdtpExceptionThrownEventsProvider.ts: Sometimes we get an invalid script id in exceptionthrown

test: https://github.com/microsoft/vscode-chrome-debug/pull/868

### DIFF
--- a/src/chrome/cdtpDebuggee/eventsProviders/cdtpExceptionThrownEventsProvider.ts
+++ b/src/chrome/cdtpDebuggee/eventsProviders/cdtpExceptionThrownEventsProvider.ts
@@ -60,15 +60,23 @@ export class CDTPExceptionThrownEventsProvider extends CDTPEventsEmitterDiagnost
     private async toExceptionDetails(exceptionDetails: CDTP.Runtime.ExceptionDetails): Promise<IExceptionDetails> {
         const scriptId = exceptionDetails.scriptId;
 
+        // We've seen scriptId contain invalid script ids. If that happens, we just ignore it
+        const scriptIfAvailable = isNotEmpty(scriptId)
+            ? await asyncUndefinedOnFailure(() => this._scriptsRegistry.getScriptByCdtpId(scriptId))
+            : undefined;
+
+        const stackTraceIfAvailable = isDefined(exceptionDetails.stackTrace)
+            ? await this._stackTraceParser.toStackTraceCodeFlow(exceptionDetails.stackTrace)
+            : undefined;
+
         return {
             exceptionId: exceptionDetails.exceptionId,
             text: exceptionDetails.text,
             lineNumber: exceptionDetails.lineNumber,
             columnNumber: exceptionDetails.columnNumber,
-            // We've seen scriptId contain invalid script ids. If that happens, we just ignore it
-            script: isNotEmpty(scriptId) ? await asyncUndefinedOnFailure(() => this._scriptsRegistry.getScriptByCdtpId(scriptId)) : undefined,
+            script: scriptIfAvailable,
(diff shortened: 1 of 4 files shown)


src/chrome/cdtpDebuggee/eventsProviders/cdtpExceptionThrownEventsProvider.ts
```diff
@@ -57,12 +58,15 @@ export class CDTPExceptionThrownEventsProvider extends CDTPEventsEmitterDiagnost
...
+        const scriptId = exceptionDetails.scriptId;
+
         return {
             exceptionId: exceptionDetails.exceptionId,
             text: exceptionDetails.text,
             lineNumber: exceptionDetails.lineNumber,
             columnNumber: exceptionDetails.columnNumber,
-            script: isNotEmpty(exceptionDetails.scriptId) ? await this._scriptsRegistry.getScriptByCdtpId(exceptionDetails.scriptId) : undefined,
+            // We've seen scriptId contain invalid script ids. If that happens, we just ignore it
+            script: isNotEmpty(scriptId) ? await asyncUndefinedOnFailure(() => this._scriptsRegistry.getScriptByCdtpId(scriptId)) : undefined,
```
EricCornelson: Suggestion: move this logic out of the declaration of the return object and create a variable. It's just a lot of stuff happening on one line in the middle of a declaration.

src/chrome/internal/exceptions/pauseOnExceptionRequestHandlers.ts
```diff
@@ -44,12 +44,14 @@ export class PauseOnExceptionRequestHandlers implements ICommandHandlerDeclarer
...
     }
 
-    private toPauseOnPromiseRejectionsStrategy(exceptionFilters: string[]): IPauseOnPromiseRejectionsStrategy {
-        if (exceptionFilters.indexOf('promise_reject') >= 0) {
-            return new PauseOnAllRejections();
-        } else {
-            return new DoNotPauseOnAnyRejections();
-        }
+    private toPauseOnPromiseRejectionsStrategy(_exceptionFilters: string[]): IPauseOnPromiseRejectionsStrategy {
+        return new PauseOnAllRejections();
```
EricCornelson: Does this mean we will pause on all rejections regardless of whether the feature is turned on or not?

src/chrome/cdtpDebuggee/eventsProviders/cdtpExceptionThrownEventsProvider.ts
digeff: Done

src/chrome/internal/exceptions/pauseOnExceptionRequestHandlers.ts
digeff: At the moment we pause on all rejections if the "Uncaught exception" option is turned on. I tested a "caught" rejection, and we also pause there, which seems a bug. I tested this on V1 and Chrome Dev Tools and they both pause on "caught" rejections too (We get a Debugger.pause message which says: "uncaught": true, so this is probably an issue with Chrome/V8)

PR #467, opened 2019-07-01
Fix attach

digeff:
- When we attach and Enable the Runtime domain we can get a Runtime.consoleAPICalled message with a script id before we enable and process all the scriptParsed messages from the Debugger domain. We know retry for a little while on getScriptByCdtpId to give time to the scriptParsed message to be processed.

- The IDebuggeeRunner has logic to detect that the user requested url was loaded. This doesn't make sense for the attach scenario, so we don't call it.

- Modified retryAsync to also retry on sync exceptions.

Test: https://github.com/microsoft/vscode-chrome-debug/pull/874

### DIFF
--- a/src/chrome/cdtpDebuggee/registries/cdtpScriptsRegistry.ts
+++ b/src/chrome/cdtpDebuggee/registries/cdtpScriptsRegistry.ts
@@ -2,7 +2,7 @@
  * Copyright (C) Microsoft Corporation. All rights reserved.
  *--------------------------------------------------------*/
 
- import { Protocol as CDTP } from 'devtools-protocol';
+import { Protocol as CDTP } from 'devtools-protocol';
 import { IScript } from '../../internal/scripts/script';
 import { ValidatedMap } from '../../collections/validatedMap';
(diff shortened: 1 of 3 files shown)


src/chrome/cdtpDebuggee/registries/cdtpScriptsRegistry.ts
```diff
@@ -110,7 +113,10 @@ class CDTPCurrentGeneration {
     }
 
     public getScriptByCdtpId(runtimeScriptCrdpId: string): Promise<IScript> {
-        return this._cdtpIdByScript.get(runtimeScriptCrdpId);
+        // When we attach to a page, we may get Runtime.consoleAPICalled messages with a reference to a ScriptId before we get
+        // the corresponding Debugger.ScriptParsed event. Wait for some time before failing this method, to give time for the message to be parsed
```
roblourens: That's unfortunate. I wonder how chrome devtools handles this.

digeff: Answered offline.

PR #469, opened 2019-07-01
Now we don't fail if we get 2 simultaneos requests

digeff:
Calling SourceTextRetriever.text with the same parameter twice simultaneously was failing because this._sourceToText.set was failing while trying to re-set the contents of a file.
Now we store the promise instead of the text, so the second call will return the promise created by the first one, and it won't fail.

### DIFF
--- a/src/chrome/internal/sources/sourceTextRetriever.ts
+++ b/src/chrome/internal/sources/sourceTextRetriever.ts
@@ -19,17 +19,18 @@ import { printArray } from '../../collections/printing';
  */
 @injectable()
 export class SourceTextRetriever {
-    private _sourceToText = new ValidatedMap<ILoadedSource, string>();
+    private _sourceToText = new ValidatedMap<ILoadedSource, Promise<string>>();
 
     constructor(@inject(TYPES.IScriptSources) private readonly _scriptSources: IScriptSourcesRetriever) { }
 
-    public async text(loadedSource: ILoadedSource): Promise<string> {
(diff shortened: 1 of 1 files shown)


src/chrome/internal/sources/sourceTextRetriever.ts
```diff
@@ -19,28 +19,29 @@ import { printArray } from '../../collections/printing';
  */
 @injectable()
 export class SourceTextRetriever {
-    private _sourceToText = new ValidatedMap<ILoadedSource, string>();
+    private _sourceToText = new ValidatedMap<ILoadedSource, Promise<string>>();
 
     constructor(@inject(TYPES.IScriptSources) private readonly _scriptSources: IScriptSourcesRetriever) { }
 
-    public async text(loadedSource: ILoadedSource): Promise<string> {
+    // We want this method to be sync, so if we get 2 simultaneous calls, the second call will return the promise of the first call
```
roblourens: This method still returns a promise, I don't think there is a behavior change here?

roblourens: Oh, now when there are two calls you will set the promise in the map instead of calling getScriptSource twice, got it. Don't call it sync though

digeff: Changed to:
    // We want this method to add an entry to the map this._sourceToText atomically, so if we get 2 simultaneous calls,
    // the second call will return the promise/result of the first call

PR #470, opened 2019-07-02
Fix setting boolean variable

digeff:
Fix needed for this test: https://github.com/microsoft/vscode-chrome-debug/pull/870

Setting a boolean was throwing the exception on line 173 before this change

### DIFF
--- a/src/chrome/chromeUtils.ts
+++ b/src/chrome/chromeUtils.ts
@@ -154,8 +154,6 @@ export function remoteObjectToValue(object: CDTP.Runtime.RemoteObject, stringify
             }
         } else if (object.type === 'undefined') {
             value = 'undefined';
-        } else if (object.type === 'boolean') {
-            value = object.value;
         } else if (object.type === 'function') {
             if (object.description === undefined) {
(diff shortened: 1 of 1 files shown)


src/chrome/chromeUtils.ts
```diff
@@ -170,7 +172,7 @@ export function remoteObjectToValue(object: CDTP.Runtime.RemoteObject, stringify
             }
         } else {
             if (object.description === undefined) {
-                throw new Error(`Expected an object that is neither objecr, not function nor undefined to have a description yet it didn't: ${JSON.stringify(object)}`);
+                throw new Error(`Expected an object that is neither object, not function nor undefined to have a description yet it didn't: ${JSON.stringify(object)}`);
             }
 
             // The value is a primitive value, or something that has a description (not object, primitive, or undefined). And force to be string
```
roblourens: I thought it was supposed to be handled by this block

roblourens: Are you sure the exception is correct? I guess that isn't in master

digeff: I missed that before. Changing implementation

roblourens:
Yeah this is basically what is in master, thanks

PR #471, opened 2019-07-04
Eng improvements

roblourens:
Is this node types update ok? VS Code just updated to 10.11.

How were you building -core? The watch task was broken for me.

@digeff @EricCornelson


EricCornelson:
I've just been building on demand with `npm run build`. Types update should be good. On that note, are we also able to update the compile target in tsconfig to ES2017? What's the story for backwards compat?

roblourens:
Yeah, we can change the build target too. I don't care about backcompat for VS Code, I will just update the min supported VS Code version to whatever we need.

PR #472, opened 2019-07-04
Fix a couple issues with electron

roblourens:
@digeff @EricCornelson 

From https://github.com/microsoft/vscode-chrome-debug/issues/878

- file:/// URIs are not handled correctly on mac
- file/url shouldn't be required at least when using runtimeExecutable


src/chrome/client/chromeDebugAdapter/cdaConfiguration.ts
```diff
@@ -53,9 +53,6 @@ export class ConnectedCDAConfiguration implements IConnectedCDAConfiguration {
             launchUrl = utils.pathToFileURL(this.args.file);
         } else if (this.args.url) {
             launchUrl = this.args.url;
-        } else {
```
digeff: Would it make more sense to change this to else if (isUndefined(config.runtimeExecutable)) instead of removing this branch?

roblourens: `runtimeExecutable` is only known about by chrome-debug. How can it participate in this validation?

digeff: You are right. 
This code should live in chrome-debug. Requested url doesn't make sense for node.
Can you add a // TODO: Move this to chrome-debug and re-enable this throw new error with else if (isUndefined(config.runtimeExecutable)) throw new Error(`You must specify either file or url to launch Chrome against a local file or a url. None were specified. ` + `The specified parameterse are: ${JSON.stringify(this.args)}`);

PR #478, opened 2019-07-12
Improved the startup error messages and other fixes

digeff:
cdtpScriptsRegistry.ts: After we clear the execution contexts, CRDP sometimes reuse ids...
chromeUtils.ts: url.parse(canonicalUrl).pathname can be null

Also improve the error messages we show if we cannot find the url that the user requested in Chrome

### DIFF
--- a/src/chrome/chromeUtils.ts
+++ b/src/chrome/chromeUtils.ts
@@ -16,7 +16,6 @@ import { IResourceIdentifier } from './internal/sources/resourceIdentifier';
 import { isNotEmpty, isEmpty, isDefined, hasMatches, isUndefined, defaultWhenEmpty } from './utils/typedOperators';
 import * as _ from 'lodash';
 import { notEmpty } from '../validation';
-import isNull = require('lodash/isNull');
 
 /**
  * Takes the path component of a target url (starting with '/') and applies pathMapping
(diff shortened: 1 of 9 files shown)


src/chrome/chromeUtils.ts
```diff
@@ -16,6 +16,7 @@ import { IResourceIdentifier } from './internal/sources/resourceIdentifier';
 import { isNotEmpty, isEmpty, isDefined, hasMatches, isUndefined, defaultWhenEmpty } from './utils/typedOperators';
 import * as _ from 'lodash';
 import { notEmpty } from '../validation';
+import isNull = require('lodash/isNull');
```
roblourens: Do we have to use `require`? It has types right?

src/chrome/client/chromeDebugAdapter/baseCDAState.ts
```diff
@@ -21,7 +21,7 @@ export abstract class BaseCDAState implements IDebugAdapterState {
         const allDeclarers = requestHandlerDeclarers.concat({
             getCommandHandlerDeclarations: () => CommandHandlerDeclaration.fromLiteralObject(requestHandlerMappingsWithDefault)
         });
-        this._requestProcessor = new RequestProcessor(allDeclarers);
+        this._requestProcessor = new RequestProcessor(`${this}`, allDeclarers);
```
roblourens: `this.toString()` would be clearer

src/chrome/chromeUtils.ts
digeff: We can use import, and that was the wrong library.
I wrote some custom code because our existing functions don't work for this particular case

src/chrome/client/chromeDebugAdapter/baseCDAState.ts
digeff: Changed

PR #480, opened 2019-07-13
Fix path transformer bug in VS

EricCornelson:
Get rid of the duplicate transformer variables and just have a single source of truth for them in _transformers. Fixes a bug in VS where the _pathTransformer gets overwritten and gets out of sync with _transformers.


EricCornelson:
> This is only an issue for VS right? that seems to be the case

Yep. Not sure what's up with the tests from Linux/Mac, looks like they've been failing since 2bfcdc020699fee737ddc5e1d445af105ffedcd2?

roblourens:
Sorry about that! This is a weird interaction with the fact that I override `path` with `path.win32` on all platforms for tests, which was probably not a good idea, and now the "properJoin" util bypasses that. I am just changing that back to path.join which is fine for VS, I'm not sure what else to do.

PR #482, opened 2019-07-13
Reduce logging

digeff:
We stop logging several methods that seem lower priority at the moment, to try to keep our logs size in check...

cdtpDebuggeeExecutionEventsProvider.ts: Avoid triggering the failure case for -1 (It happens all the time)
printObjectDescription.ts: print "a ClassName" when an object toString prints [object object]
chromeDebugSession.ts: Now From client: and To  client: will be aligned with the START, END, SNRET lines

### DIFF
--- a/src/chrome/logging/methodsCalledLogger.ts
+++ b/src/chrome/logging/methodsCalledLogger.ts
@@ -64,7 +64,7 @@ export class MethodsCalledLogger<T extends object> {
         const handler = {
             get: <K extends keyof T>(target: T, propertyKey: K, receiver: any) => {
                 const originalPropertyValue = target[propertyKey];
-                if (!shouldLog(originalPropertyValue)) {
+                if (!shouldLog(originalPropertyValue, propertyKey)) {
                     return originalPropertyValue;
                 }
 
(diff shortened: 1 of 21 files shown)


src/chrome/logging/methodsCalledLogger.ts
```diff
@@ -128,17 +132,17 @@ export class MethodsCalledLogger<T extends object> {
...
     private logCallStart(propertyKey: PropertyKey, methodCallArguments: any[], callId: number): void {
-        const message = `START            ${callId}: ${this.printMethodCall(propertyKey, methodCallArguments)}`;
+        const message = `START ${callId}: ${this.printMethodCall(propertyKey, methodCallArguments)}`;
         logger.verbose(message);
     }
 
-    private logSyncPartFinished(propertyKey: PropertyKey, methodCallArguments: any[], callId: number): void {
-        const message = `PROMISE-RETURNED ${callId}: ${this.printMethodCall(propertyKey, methodCallArguments)}`;
+    private logSyncPartFinished(propertyKey: PropertyKey, _methodCallArguments: any[], callId: number): void {
+        const message = `SNRET ${callId}: ${this.printMethodName(propertyKey)}`;
```
EricCornelson: what does SNRET mean?

digeff: It's my abbreviation of synchronic return

PR #483, opened 2019-07-15
Send telemetry for errors

digeff:
Send telemetry for errors


src/chrome/utils/failures.ts
```diff
@@ -1,3 +1,5 @@
+import { telemetry } from '../../telemetry';
```
roblourens: This should go below the copyright (autoimport doesn't know that...)

digeff: Done

PR #484, opened 2019-07-16
Ignore debugger internal sources

digeff:
When we enable ReportVersionInformation we evaluate navigator.userAgent on the debuggee which creates a script which makes our loaded tests fail because we get unexpected loaded sources.
We shouldn't send loaded sources for the internal scripts used by the debug adapter


digeff:
The best we have at the moment is the isV1 flag on the tests, that will show how V1 vs V2 differ.

PR #487, opened 2019-07-17
v2 version number

digeff:
Real version number for v2, so we can start writing telemetry queries for real...


EricCornelson:
Why 7.0.0?

digeff:
v1 is 6.7.52

EricCornelson:
Got it, I was looking at the 10.7.52 and forgot that that was a spoof

PR #491, opened 2019-07-23
Fix logging

digeff:
We weren't using the logging level of the trace parameter


src/chrome/client/chromeDebugAdapter/unconnectedCDA.ts
```diff
@@ -60,11 +60,22 @@ export class UnconnectedCDA implements IDebugAdapterState {
     private parseLoggingConfiguration(args: ILaunchRequestArgs | IAttachRequestArgs): ILoggingConfiguration {
         let traceValue: Logger.LogLevel;
         switch (args.trace) {
+            case true:
+            case 'true':
```
roblourens: Why string `"true"`?

digeff: In case someone mistakenly sends "true" instead of true.

roblourens: Well they should not do that...

digeff: Removing it.

PR #492, opened 2019-07-24
Log VS Code protocol as info level

digeff:
We were most some To client: and From client: logs messages on the "info" level.
We want to keep them.

We modify the logger.verbose to keep only those messages


digeff:
I increased the version number so we can distinguish it in telemetry form the previous version

PR #494, opened 2019-07-25
Localization

digeff:
Added localize call to all the strings that might be shown to the user

(Most errors can produce a failed request, which will be sent as the message of a DebugProtocol.response which can be shown to the user)

### DIFF
--- a/src/chrome/chromeDebugAdapter.ts
+++ b/src/chrome/chromeDebugAdapter.ts
@@ -499,7 +499,7 @@ export class ChromeDebugLogic {
             }
 
             if (isDefined(response.exceptionDetails)) {
-                // Not an error, getter could be `get foo() { throw new Error('bar'); }`
+                // Not an error, getter could be `get foo() { throw new Error(localize('error.bar', 'bar')); }`
                 const exceptionMessage = ChromeUtils.errorMessageFromExceptionDetails(response.exceptionDetails);
                 logger.log('Exception thrown evaluating getter - ' + exceptionMessage);
                 return { name: propDesc.name, value: exceptionMessage, variablesReference: 0 };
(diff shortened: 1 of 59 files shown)


src/chrome/chromeDebugAdapter.ts
```diff
@@ -643,7 +643,7 @@ export class ChromeDebugLogic {
     }
 
     protected threadName(): string {
-        return 'Thread ' + ChromeDebugLogic.THREAD_ID;
+        return localize('thread.prefix', 'Thread ') + ChromeDebugLogic.THREAD_ID;
```
roblourens: Space should go outside of the localized string so it isn't lost in localization

roblourens: Actually the id should be an arugument like {0} in case it goes before "Thread" in some language

digeff: Changed to return localize('thread.prefix', 'Thread {0}', ChromeDebugLogic.THREAD_ID);

src/chrome/internal/breakpoints/features/bpsDeltaCalculator.ts
```diff
@@ -89,8 +89,8 @@ export class BPRsDeltaCalculator {
                 existingToLeaveAsIs: this.printLocations(match.existingToLeaveAsIs)
             };
 
-            const additionalDetails = localize('\nRequested breakpoints = {0}\nExisting breakpoints = ${1}\nMatch = ${2}', JSON.stringify(this._requestedBPRecipes.breakpoints.map(canonicalizeBPLocation)), JSON.stringify(Array.from(this._currentBPRecipes).map(canonicalizeBPLocation)), JSON.stringify(matchJson));
-            throw new Error(localize('error.deltaCalculator.deltaCalculatorInvalidResult', '{0}\nmatch: {1}', errorMessage, additionalDetails));
+            const additionalDetails = localize('error.deltaCalculator.invalidResultDetails.', '\nRequested breakpoints = {0}\nExisting breakpoints = ${1}\nMatch = ${2}', JSON.stringify(this._requestedBPRecipes.breakpoints.map(canonicalizeBPLocation)), JSON.stringify(Array.from(this._currentBPRecipes).map(canonicalizeBPLocation)), JSON.stringify(matchJson));
```
roblourens: `.` in string id

digeff: Fixed

digeff:
Will re-open when it's ready to be merged

PR #495, opened 2019-07-26
Perform column detection just once

digeff:
We were performing the column detection and calling getPossibleBreakpoints once per script. Now we perform it only once.


roblourens:
The windows integration tests are still failing

PR #503, opened 2019-07-30
Localization

digeff:
Added localize call to all the strings that might be shown to the user, and internal errors to the ones that aren't actionable

(Most errors can produce a failed request, which will be sent as the message of a DebugProtocol.response which can be shown to the user)

TODO: Figure out what is the right thing to do for ErrorWithMessage in errors.ts


src/chrome/cdtpDebuggee/features/notes.json
```diff
@@ -0,0 +1,10 @@
+{
```
roblourens: Is this needed?

digeff: Whoops... Didn't mean to commit this file... Thanks for flagging this :)

PR #506, opened 2019-08-05
Make localization work for v2, and send locale independent error ids

digeff:
This PR solves 2 different issues:
 - It makes localization work for v2 by re-creating the localize function after we change the locale
 - It provides a LocalizedError class where you can configure the error id for a localized error message (This will let us deduplicate errors across different locales in telemetry)

src/chrome/utils/localizedError.ts implements the solution for both of those issues

### DIFF
--- a/src/chrome/utils/localizedError.ts
+++ b/src/chrome/utils/localizedError.ts
@@ -3,40 +3,6 @@
  *--------------------------------------------------------*/
 
 import * as nls from 'vscode-nls';
-import { LocalizeInfo } from 'vscode-nls';
-import _ = require('lodash');
-
-const originalConfig = nls.config;
-let currentLocale: string | undefined = undefined;
-
-/** We replace nls.config with our own version so we can know which locale is set, and force all the localize functions to be re-created when needed.
- * This works around the need to import all the secondary files only after the locale has been already set.
- */
-(<any>nls).config = (opts?: nls.Options) => {
-    currentLocale = _.defaultTo(opts, { locale: undefined }).locale;
-    return originalConfig(opts);
-};
-
-const originalLoadMessageBundle = nls.loadMessageBundle;
-/** We replace nls.loadMessageBundle with our own custom version, which will re-created the localize function after the locale has been changed.
- * This works around the need to import all the secondary files only after the locale has been already set.
- */
-(<any>nls).loadMessageBundle = (file?: string) => {
-    let localizeForFile = originalLoadMessageBundle(file);
-    let localizeForFileLocale = currentLocale;
-    return (key: string | LocalizeInfo, message: string, ...args: (string | number | boolean | undefined | null)[]) =>
-        {
-            // Has the locale changed since we created the localize function?
-            if (localizeForFileLocale !== currentLocale) {
-                localizeForFileLocale = currentLocale;
-                // If so, re-create it so it'll use the new local (the localize caches the local, so we do need to re-create it after changing the locale)
-                localizeForFile = originalLoadMessageBundle(file);
-            }
-
-            // Call our custom localize which will store the localization key for the LocaliedErrors
-            return localizeForFile(<string>key, message, ...args);
-        };
-};
 
 /** Error class which automatically extracts the localization key of the latest call to localize (So we can easily deduplicate errors on telemetry) */
 export class LocalizedError extends Error {
@@ -44,4 +10,15 @@ export class LocalizedError extends Error {
     public constructor(public readonly errorCode: string, localizedMessage: string) {
         super(localizedMessage);
     }
+}
+
+const getLocalizeCalls: (() => nls.LocalizeFunc)[] = [];
+
+export function registerGetLocalize(getLocalizeCallback: () => nls.LocalizeFunc): void {
+    getLocalizeCalls.push(getLocalizeCallback); // Add to the list to replace with the correct locale after it gets configured
+}
+
+export function configureLocale(locale: string): void {
+    nls.config({ locale: locale });
+    getLocalizeCalls.forEach(eachCall => eachCall());
 }
--- a/src/chrome/utils/localizedError.ts
+++ b/src/chrome/utils/localizedError.ts
@@ -1,0 +1,119 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+import * as nls from 'vscode-nls';
+import * as path from 'path';
+import { LocalizeInfo, LocalizeFunc } from 'vscode-nls';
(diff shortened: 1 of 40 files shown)


src/chrome/utils/localizedError.ts
```diff
@@ -0,0 +1,119 @@
...
+// Consume the key of the latest call to the localize function
+function consumeLastLocalizationKey(): string {
+    const lastKeyToReturn = lastKeyOfLocalize;
+    lastKeyOfLocalize = defaultLastKeyOfLocalize;
+    return lastKeyToReturn;
+}
+
+/** Error class which automatically extracts the localization key of the latest call to localize (So we can easily deduplicate errors on telemetry) */
+export class LocalizedError extends Error {
+    public readonly errorCode = consumeLastLocalizationKey();
```
roblourens: This is not a safe assumption, how can you know that localize will always be called immediately before creating a LocalizedError?

src/chrome/utils/localizedError.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/** Format of the bundle.json file */
+interface LocalizationBundle {
+    [file: string]: { keys: (string | undefined)[] } | undefined;
+}
+
+/** We store the bundle metadata on memory to be able to extract the localization keys based on the file and message index */
+let bundle: LocalizationBundle = {};
+const bundlePath = path.resolve(__dirname, '../../../nls.metadata.json');
```
roblourens: This is not safe, the shape of this file and the transformations that vscode-nls/vscode-nls-dev do are not part of their API, they are internal implementation details. They have changed in the past and can change again.

src/chrome/utils/localizedError.ts
```diff
@@ -0,0 +1,119 @@
...
+(<any>nls).config = (opts?: nls.Options) => {
+    currentLocale = _.defaultTo(opts, { locale: undefined }).locale;
+    return originalConfig(opts);
+};
+
+const originalLoadMessageBundle = nls.loadMessageBundle;
+/** We replace nls.loadMessageBundle with our own custom version, which will re-created the localize function after the locale has been changed.
+ * This works around the need to import all the secondary files only after the locale has been already set.
+ */
+(<any>nls).loadMessageBundle = (file?: string) => {
```
roblourens: I don't understand this either

src/chrome/utils/localizedError.ts
```diff
@@ -0,0 +1,119 @@
...
+
+// Consume the key of the latest call to the localize function
+function consumeLastLocalizationKey(): string {
+    const lastKeyToReturn = lastKeyOfLocalize;
+    lastKeyOfLocalize = defaultLastKeyOfLocalize;
+    return lastKeyToReturn;
+}
+
+/** Error class which automatically extracts the localization key of the latest call to localize (So we can easily deduplicate errors on telemetry) */
+export class LocalizedError extends Error {
```
roblourens: I'm not sure the error code should be tied to the localized string id at all. These are two different things and it seems like they should operate independantly. Maybe we will want to change the error code without losing the localized string, or have multiple errors share the same string, or multiple strings associated with one error, or have an error code without a string attached, etc etc. Plus this behavior is hidden and non-obvious.

src/chrome/utils/localizedError.ts
digeff: There is no warranty. At the moment all the calls of new LocalizedError( have localize as it's first argument, so it should be true.
If this where ever to be false we would return "last localization key couldn't be determined" and see that in the telemetry, and we could investigate more to figure out why that happened.

src/chrome/utils/localizedError.ts
digeff: That's a good point. Changing to a different approach so we don't tied the error-id and the localization-id.

src/chrome/utils/localizedError.ts
digeff: That's a good point. Changing to a different approach so we don't rely on the format of that file.

src/chrome/utils/localizedError.ts
digeff: From what I've experimented of nls, when we call 
nls.loadMessageBundle() on a file the first time, it returns a localize function that works in English.

After we call nls.config on the initialize message, all the localize functions are already created, and tied to the English language, so the translations are not being used. To get the translations to be used, we need to call nls.loadMessageBundle() for all files again, to get a new version of localize that is tied to the proper locale.

This methods overrides nls.loadMessageBundle() to automatically re-create the localize function after a locale is specified in nls.config

roblourens: Like this works in v1. So why are weird hacks needed? If there is a bug in the library let's fix that.

src/chrome/utils/localizedError.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    public constructor(public readonly errorCode: string, localizedMessage: string) {
+        super(localizedMessage);
+    }
+}
+
+const getLocalizeCalls: (() => nls.LocalizeFunc)[] = [];
+
+export function registerGetLocalize(getLocalizeCallback: () => nls.LocalizeFunc): void {
```
roblourens: Please give a couple sentences about why this is needed, and what happens when you call localize before and after this callback is called, so we don't forget :)

roblourens: Also I don't think these should be in a file called "localizeError.ts"

src/chrome/variables.ts
```diff
@@ -16,6 +15,10 @@ import { CDTPNonPrimitiveRemoteObject, validateNonPrimitiveRemoteObject } from '
 import { isDefined, isUndefined, hasMatches, isNotEmpty, isTrue } from './utils/typedOperators';
 import * as _ from 'lodash';
 import { InternalError } from './utils/internalError';
+import { registerGetLocalize } from './utils/localizedError';
+
+let localize = nls.loadMessageBundle();
```
roblourens: I guess it's too bad that all bundles would be loaded twice, but that's ok.

src/chrome/utils/localizedError.ts
digeff: Renamed to localization.ts
Added comment:
/** Each file that calls localize = nls.loadMessageBundle() needs to call registerGetLocalize(() => localize = nls.loadMessageBundle()); too.
 * We call localize = nls.loadMessageBundle() when the file is parsed, which in most cases is before the DAP.initializeRequest which specifies the locale we need to
 * use in Visual Studio (In VS Code this works because it passes the locale in an environment variable).
 * So after we get the locale from Visual Studio, we need to re-create all the localize functions in all the program to use the right locale, so we use
 * registerGetLocalize to register all those places, and have all those locale functions recreated
 */

PR #508, opened 2019-08-08
Adding support for launching with port = 0

EricCornelson:
In order to fix some of the timeout issues we have in VS, we want to be able to support launching with --remote-debugging-port=0 on Chrome, which will then allow Chrome to select the debugging port, and let us know through the DevToolsActivePort file in the user-data-dir.

### DIFF
--- a/src/chrome/chromeConnection.ts
+++ b/src/chrome/chromeConnection.ts
@@ -95,7 +95,7 @@ export interface IChromeError {
  * Connects to a target supporting the Chrome Debug Protocol and sends and receives messages
  */
 export class ChromeConnection implements IObservableEvents<IStepStartedEventsEmitter> {
-    private static ATTACH_TIMEOUT = 10000; // ms
+    protected static ATTACH_TIMEOUT = 10000; // ms
 
     private _socket: WebSocket;
     private _crdpSocketMultiplexor: CRDPMultiplexor;
--- a/src/chrome/chromeUtils.ts
+++ b/src/chrome/chromeUtils.ts
@@ -12,6 +12,8 @@ import { ITarget } from './chromeConnection';
 import { IPathMapping } from '../debugAdapterInterfaces';
 import { pathToRegex } from '../utils';
 import { isInternalRemotePath } from '../remoteMapper';
+import { Socket } from 'net';
+import * as errors from '../errors';
 
 /**
  * Takes the path component of a target url (starting with '/') and applies pathMapping
@@ -311,4 +313,49 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
     const fileNameWithoutExtension = path.parse(fileNameWithoutFullPath).name;
     const escapedFileName = pathToRegex(fileNameWithoutExtension);
     return '.*[\\\\\\/]' + escapedFileName + '([^A-z^0-9].*)?$';
+}
+
+/**
+ * Checks if a given tcp port is currently in use (more accurately, is there a server socket accepting connections on that port)
+ * @param port The port to check
+ * @param host Optional host, defaults to 127.0.0.1
+ * @param timeout Timeout for the socket connect attempt
+ * @returns True if a server socket is listening on the given port, false otherwise
+ */
+export async function isPortInUse(port: number, host = '127.0.0.1', timeout = 400): Promise<boolean> {
+    // Basically just create a socket and try to connect on that port, if we can connect, it's open
+    return new Promise<boolean>((resolve, _reject) => {
+        const socket = new Socket();
+        function createCallback(inUse: boolean) {
+            return () => {
+                resolve(inUse);
+                socket.removeAllListeners();
+                socket.destroy();
+            };
+        }
+
+        socket.setTimeout(timeout);
+        socket.on('connect', createCallback(true));
+        socket.on('timeout', createCallback(false));
+        socket.on('error', createCallback(false));
+        socket.connect(port, host);
+    });
+}
+
+/**
+ * Get the port on which chrome was launched when passed "--remote-debugging-port=0"
+ * @param userDataDir The profile data directory for the Chrome instance to check
+ * @throws If reading the port failed for any reason
+ */
+export async function getLaunchedPort(userDataDir: string): Promise<number> {
+    const activePortFilePath = path.join(userDataDir, 'DevToolsActivePort');
+    try {
+        const activePortArgs = await utils.readFileP(activePortFilePath, 'utf-8');
+        const [ portStr ] = activePortArgs.split('\n'); // chrome uses \n regardless of platform in this file
+        const port = parseInt(portStr, 10);
+        if (isNaN(port)) return Promise.reject(errors.activePortFileContentsInvalid(activePortFilePath, activePortArgs));
+        return port;
+    } catch (err) {
+        return Promise.reject(errors.failedToReadPortFromUserDataDir(userDataDir, err));
+    }
 }
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -151,3 +151,21 @@ export function noStoredException(): DebugProtocol.Message {
         sendTelemetry: true
     });
 }
+
+export function failedToReadPortFromUserDataDir(dataDirPath: string, err: Error) {
+    return new ErrorWithMessage({
+        id: 2033,
+        format: localize('failed.to.read.port', 'Failed to read file {dataDirPath}, {error}'),
+        variables: { dataDirPath, error: err.message },
+        sendTelemetry: true
+    });
+}
+
+export function activePortFileContentsInvalid(dataDirPath: string, dataDirContents: string) {
+    return new ErrorWithMessage({
+        id: 2034,
+        format: localize('port.file.contents.invalid', 'File at location: "{dataDirPath}" did not contain valid port data, contents were: {dataDirContents}'),
+        variables: { dataDirPath, dataDirContents },
+        sendTelemetry: true
+    });
+}
--- a/test/testUtils.ts
+++ b/test/testUtils.ts
@@ -107,7 +107,8 @@ export function pathResolve(...segments: string[]): string {
 }
 
 export function registerMockReadFile(...entries: { absPath: string; data: string }[]): void {
-    const fsMock = Mock.ofInstance(fs, MockBehavior.Strict);
+    const fsMock = Mock.ofInstance(fs, MockBehavior.Loose);
+    fsMock.callBase = true; // use real fs if we didn't mock the call
     mockery.registerMock('fs', fsMock.object);
 
     entries.forEach(entry => {


src/chrome/chromeUtils.ts
```diff
@@ -312,3 +315,45 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
...
 }
+
+/**
+ * Checks if a given tcp port is currently in use (more accurately, is there a server socket accepting connections on that port)
+ * @param port The port to check
+ * @param host Optional host, defaults to 127.0.0.1
+ * @param timeout Timeout for the socket connect attempt
+ * @returns True if a server socket is listening on the given port, false otherwise
+ */
+export async function isPortInUse(port: number, host: string = '127.0.0.1', timeout = 400): Promise<boolean> {
```
digeff: Would it be better to use localhost as a default?
Is it possible that some customer is running this on a PC with only IPv6?

src/chrome/chromeUtils.ts
```diff
@@ -312,3 +315,45 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
...
+}
+
+/**
+ * Get the port on which chrome was launched when passed "--remote-debugging-port=0"
+ * @param userDataDir The profile data directory for the Chrome instance to check
+ * @throws If reading the port failed for any reason
+ */
+export async function getLaunchedPort(userDataDir: string): Promise<number> {
+    let port = 0;
+    const activePortFilePath = path.join(userDataDir as any, 'DevToolsActivePort');
```
digeff: why do you need to cast to any?

src/chrome/chromeUtils.ts
```diff
@@ -312,3 +315,45 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
...
+    const activePortArgs = (await promisify(fs.readFile)(activePortFilePath)).toString();
```
digeff: There is some readFileP around that does this same thing

src/chrome/chromeUtils.ts
```diff
@@ -312,3 +315,45 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
...
+    const [ portStr ] = activePortArgs.split('\n'); //chrome uses \n regardless of platform in this file
+    if(portStr) {
+        port = parseInt(portStr);
+    }
+    if(port === NaN) throw new Error(`Reading ${portStr} from ${activePortArgs} returned NaN`);
```
digeff: if portStr  is "" you'll return 0
We should also send telemetry if the file doesn't have the format we expect.

src/chrome/chromeConnection.ts
```diff
@@ -192,4 +191,22 @@ export class ChromeConnection implements IObservableEvents<IStepStartedEventsEmi
...
+    /**
+     * Gets the port on which chrome was launched, and throw error if the port is not open or accepting connections
+     * @param host The host address on which to check if the port is listening
+     * @param userDataDir Chrome user data directory in which to check for a port file
+     */
+    private async getLaunchedPort(host: string, userDataDir: string): Promise<number> {
+        logger.verbose('Looking for DevToolsActivePort file...');
+        const launchedPort = await chromeUtils.getLaunchedPort(userDataDir);
+        logger.verbose('Got the port, checking if its ready...');
+        const portInUse = await chromeUtils.isPortInUse(launchedPort, host, 100);
```
digeff: Do we need to verify that  the port is in use? Would it work if we just call this._targetDiscoveryStrategy.getTarget and let it fail and retry if the port is not in use?

src/chrome/chromeConnection.ts
```diff
@@ -128,9 +129,19 @@ export class ChromeConnection implements IObservableEvents<IStepStartedEventsEmi
     /**
      * Attach the websocket to the first available tab in the chrome instance with the given remote debugging port number.
      */
-    public attach(address = '127.0.0.1', port = 9222, targetUrl?: string, timeout?: number, extraCRDPChannelPort?: number): Promise<void> {
-        return this._attach(address, port, targetUrl, timeout, extraCRDPChannelPort)
-            .then(() => { });
+    public attach(address = '127.0.0.1', port = 9222, targetUrl?: string, timeout = ChromeConnection.ATTACH_TIMEOUT, extraCRDPChannelPort?: number, userDataDir?: string): Promise<void> {
```
digeff: Is there any way to only implement this functionality for Chrome?
This doesn't make sense for Node.js

src/chrome/chromeUtils.ts
EricCornelson: Possibly. I recall Michael mentioning that 'localhost' actually had issues on some machines, and that using '127.0.0.1' explicitly solved it. In any case, we're using the default of '127.0.0.1' in most other methods related to launching, which is why I stuck with that convention.

src/chrome/chromeConnection.ts
EricCornelson: True, but it's already optional, right? Will node ever send --user-data-dir?

src/chrome/chromeUtils.ts
EricCornelson: parseInt("") returns NaN, so we'll throw an error. I agree about telemetry, though. What's the best way to add that?

src/chrome/chromeUtils.ts
EricCornelson: Changed

src/chrome/chromeUtils.ts
EricCornelson: We don't :)

src/chrome/chromeConnection.ts
EricCornelson: We could do that, but the reason I chose to test the port explicitly was A) we know that if nothing is listening on that port, we will never be able to succesfully get targets anyway, and B) it seemed difficult to adjust the default timeout specifically for the socket connect portion of the HTTP request (which is 2s by default), this meant that often we were actually waiting 4s  to connect instead of 2.2 or whatever. We could just use promiseTimeout from our utils on that and throw away the result, but that seemed a little messy to me.

src/chrome/chromeUtils.ts
EricCornelson: Spoke offline. This is fixed.

src/chrome/chromeConnection.ts
```diff
@@ -128,9 +129,19 @@ export class ChromeConnection implements IObservableEvents<IStepStartedEventsEmi
...
+        let selectedTarget: ITarget;
+        return utils.retryAsync(async () => {
+            const launchedPort = (port === 0 && userDataDir) ? await this.getLaunchedPort(address, userDataDir) : port;
```
digeff: What will happen if someone calls this with port: 0 and no userDataDir?

src/chrome/chromeConnection.ts
```diff
@@ -192,4 +191,22 @@ export class ChromeConnection implements IObservableEvents<IStepStartedEventsEmi
...
+     */
+    private async getLaunchedPort(host: string, userDataDir: string): Promise<number> {
+        logger.verbose('Looking for DevToolsActivePort file...');
+        const launchedPort = await chromeUtils.getLaunchedPort(userDataDir);
+        logger.verbose('Got the port, checking if its ready...');
+        const portInUse = await chromeUtils.isPortInUse(launchedPort, host, 100);
+        if(!portInUse) {
+            // bail, the port isn't open
+            logger.verbose('Port not open yet...');
+            throw new Error('Port not open yet');
```
digeff: We should localize the message, and mention which port we were trying to check

src/chrome/chromeUtils.ts
```diff
@@ -312,3 +313,45 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
...
+/**
+ * Checks if a given tcp port is currently in use (more accurately, is there a server socket accepting connections on that port)
+ * @param port The port to check
+ * @param host Optional host, defaults to 127.0.0.1
+ * @param timeout Timeout for the socket connect attempt
+ * @returns True if a server socket is listening on the given port, false otherwise
+ */
+export async function isPortInUse(port: number, host: string = '127.0.0.1', timeout = 400): Promise<boolean> {
+    // Basically just create a socket and try to connect on that port, if we can connect, it's open
+    return new Promise<boolean>((accept, _reject) => {
```
digeff: I'd rename accept to resolve which is the standard name for that argument

src/chrome/chromeUtils.ts
```diff
@@ -312,3 +313,45 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
...
+ * Get the port on which chrome was launched when passed "--remote-debugging-port=0"
+ * @param userDataDir The profile data directory for the Chrome instance to check
+ * @throws If reading the port failed for any reason
+ */
+export async function getLaunchedPort(userDataDir: string): Promise<number> {
+    let port = -1;
+    const activePortFilePath = path.join(userDataDir, 'DevToolsActivePort');
+    const activePortArgs = await utils.readFileP(activePortFilePath, 'utf-8');
+    const [ portStr ] = activePortArgs.split('\n'); //chrome uses \n regardless of platform in this file
+    if(portStr) {
```
digeff: Is it the same if we call port = parseInt(portStr); no matter what, and we eliminate the     if(portStr) ?

src/chrome/chromeConnection.ts
EricCornelson: Changed this. We will now throw an error in that case. We _could_ theoretically try to be smart and figure out the userDataDir, but there are so many different possibilities I'd rather not. The only time this would ever be a problem anyway is if the user passes runtimeExecutable with no data dir and port 0.

src/chrome/chromeConnection.ts
EricCornelson: Done

src/chrome/chromeUtils.ts
EricCornelson: Done

src/chrome/chromeUtils.ts
EricCornelson: Yep, that's better, changed.

src/chrome/chromeConnection.ts
dhanvikapila: Chrome does specify defaults for the user data dir [here](https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md#windows).

src/chrome/chromeUtils.ts
roblourens: Right, don't use localhost

roblourens:
Why is this going into master and not v2?

EricCornelson:
We also want to put this into V2, but we want to test them independently (as an example, if this change introduces some weird bug and breaks a bunch of stuff, we won't immediately attribute that to v2). We'll probably merge this into V2 after this PR is approved, but not use it until we know it works as expected and fixes our VS-specific timeout issues.

src/chrome/chromeConnection.ts
EricCornelson: resolved with refactoring anyway

roblourens:
Ok that's fine. But this isn't going to ship in VS before v2 does right? So "testing" will just mean running tests/selfhosting?

roblourens:
Also the tests are failing

EricCornelson:
They may end up going out around the same time. The plan is to enable v2 (and this change) behind an experiment so we can gather some data before we release to everyone.

Tests are passing now, also.

src/chrome/chromeUtils.ts
```diff
@@ -312,3 +314,46 @@ export function getUrlRegexForBreakOnLoad(url: string): string {
...
+    return new Promise<boolean>((resolve, _reject) => {
+        const socket = new Socket();
+        const callback = (inUse: boolean) => () => {
+            resolve(inUse);
+            socket.removeAllListeners();
+            socket.destroy();
+        };
+
+        socket.setTimeout(timeout);
+        socket.on('connect', callback(true));
```
roblourens: Won't this call callback immediately, not on the event?

EricCornelson: Sorry, I realize it wasn't super clear that the `callback` function was actually creating a new function., and _that_ function was actually being passed as the callback. Changed to be clearer.

roblourens: Oh I didn't read closely. Thanks.

PR #513, opened 2019-08-23
Added PossiblyCustomerContent<T> class to prevent logging customer content inadvertently

digeff:
Updated the code to store customer content in the CustomerContent class. The most important part of this change is that JSON.stringify(new CustomerContent()) won't contain the customer CustomerContent

Replaced several uses of string for a specifically designed class. Also moved into those classes some code that processed the possibly customer data:
    * SourceMapUrl
    * SourceContents
    * SourceMapContents

### DIFF
--- a/src/chrome/internal/sources/sourceContents.ts
+++ b/src/chrome/internal/sources/sourceContents.ts
@@ -1,0 +1,28 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+ import { PossiblyCustomerContent, CustomerContent, NonCustomerContent } from '../../logging/gdpr';
+
+export class SourceContents {
+    public constructor(public readonly contents: PossiblyCustomerContent<string>) {}
+
+    public static customerContent(contents: string): SourceContents {
+        return new SourceContents(new CustomerContent(contents));
+    }
+
+    public static nonCustomerContent(contents: string): SourceContents {
+        return new SourceContents(new NonCustomerContent(contents));
+    }
+
+    public get customerContentData(): string {
+        return this.contents.customerContentData;
+    }
+
+    public truncated(): SourceContents {
+        const maxLength = 1e5;
+        return this.customerContentData.length > maxLength
+            ? SourceContents.customerContent(this.customerContentData.substr(0, maxLength) + '[⋯]')
+            : this;
+    }
+}
--- a/src/chrome/logging/gdpr.ts
+++ b/src/chrome/logging/gdpr.ts
@@ -1,0 +1,56 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+/** Used to store data that can potentially be customer content. If it is, we should never log it or send it over telemetry */
+ export interface PossiblyCustomerContent<T> {
+    customerContentData: T;
+
+    /** Transform the data. If it is customer data, the result will sill be customer data */
+    transform<U>(transformFunction: (customerContentData: T) => U): PossiblyCustomerContent<U>;
+}
+
+/** Used to protect CustomerContent data so we don't inadvertently log it or send it as telemetry */
+export class CustomerContent<T> implements PossiblyCustomerContent<T> {
+    public constructor(data: T) {
+        Object.defineProperty(this, 'customerContentData', {
+            get: () => data
+        });
+    }
+
+    public static maybeNull<T>(data: T | null): PossiblyCustomerContent<T> | null {
+        return data !== null
+            ? new CustomerContent(data)
+            : null;
+    }
+
+    public static maybeUndefined<T>(data: T | undefined): PossiblyCustomerContent<T> | undefined {
+        return data !== undefined
+            ? new CustomerContent(data)
+            : undefined;
+    }
+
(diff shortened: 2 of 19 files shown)


src/chrome/logging/gdpr.ts
```diff
@@ -0,0 +1,56 @@
...
+        });
+    }
+
+    public static maybeNull<T>(data: T | null): PossiblyCustomerContent<T> | null {
+        return data !== null
+            ? new CustomerContent(data)
+            : null;
+    }
+
+    public static maybeUndefined<T>(data: T | undefined): PossiblyCustomerContent<T> | undefined {
```
EricCornelson: both `maybeNull` and `maybeUndefined` would be more reusable as higher order functions e.g.:

```ts
function mapOrUndefined<T, R>(mappingFunc: (item: T) => R): (item: T) => R | undefined {
    return (item: T) => (item !== undefined) ? mappingFunc(item) : undefined;
}
```

which you could then call like,

```ts
    const myVar = mapOrUndefined(x => new CustomerContent(x))(myString);
```

or, if you wanted to make the convenience function specific as well,

```ts
    export const customerContentOrUndefined( content: string ) = mapOrUndefined(x => new CustomerContent(x);
```

src/sourceMaps/sourceMapUrl.ts
```diff
@@ -0,0 +1,57 @@
...
+export class SourceMapUrl {
+    private readonly _url: PossiblyCustomerContent<string>;
+    public readonly isInlineSourceMap: boolean;
+
+    public constructor(url: string) {
+        this.isInlineSourceMap = url.indexOf('data:application/json') >= 0;
+        this._url = this.isInlineSourceMap ? new CustomerContent(url) : new NonCustomerContent(url);
+    }
+
+    public static maybeUndefined(url: string | undefined): SourceMapUrl | undefined {
```
EricCornelson: duplicated code as mentioned above

src/sourceMaps/sourceMapUrl.ts
```diff
@@ -0,0 +1,57 @@
...
+ import { PossiblyCustomerContent, CustomerContent, NonCustomerContent } from '../chrome/logging/gdpr';
+import { logger } from 'vscode-debugadapter';
+import { SourceMapContents } from './sourceMapContents';
+
+export class SourceMapUrl {
+    private readonly _url: PossiblyCustomerContent<string>;
+    public readonly isInlineSourceMap: boolean;
+
+    public constructor(url: string) {
+        this.isInlineSourceMap = url.indexOf('data:application/json') >= 0;
```
EricCornelson: Is this the only time we'd consider the url customer content? I can think of other times where potentially PII could be passed in, for example, url params.

src/chrome/logging/gdpr.ts
digeff: I didn't realize this. We are not using those 2 any more. Deleted them.

src/sourceMaps/sourceMapUrl.ts
digeff: You are right, those can also be potentially customer content. For the purpose of this PR we are focusing on source code only.

src/chrome/internal/sources/sourceContents.ts
```diff
@@ -0,0 +1,28 @@
+/*---------------------------------------------------------
+ * Copyright (C) Microsoft Corporation. All rights reserved.
+ *--------------------------------------------------------*/
+
+ import { PossiblyCustomerContent, CustomerContent, NonCustomerContent } from '../../logging/gdpr';
+
+export class SourceContents {
```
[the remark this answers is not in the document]
digeff: The main purpose of the class is to provide a more specific name for the strings we are passing around, and improve the readability of the code. Which one do you think is better?

A. public async getScriptSource(script: IScript): Promise\<SourceContents> {

B. public async getScriptSource(script: IScript): Promise<PossiblyCustomerContent\<string>> {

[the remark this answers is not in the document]
EricCornelson: In that case, you could just use a type def and get the same naming benefits:
```ts
    type SourceContents = PossiblyCustomerContent<string>;
```

src/sourceMaps/sourceMapUrl.ts
EricCornelson: got it

src/chrome/logging/gdpr.ts
```diff
@@ -0,0 +1,56 @@
...
+        return data !== undefined
+            ? new CustomerContent(data)
+            : undefined;
+    }
+
+    public static fromAsync<T>(promisedData: Promise<T>): Promise<PossiblyCustomerContent<T>> {
```
EricCornelson: This could also be made into a utility function

src/sourceMaps/sourceMapUrl.ts
digeff: I changed the code, it's not duplicated any more

src/chrome/logging/gdpr.ts
digeff: I used to have 3 callers of this. I only have 1 now. Removing this utility function

src/chrome/internal/sources/sourceContents.ts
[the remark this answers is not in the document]
digeff: Changed!

roblourens:
Build broken?

PR #515, opened 2019-08-27
Add new step to startup timing telemetry

mrcrane:
to make attach measurements more accurate


mrcrane:
There seems to be a flaky unit test, but it's all green now

Issue #514, opened 2019-08-25
Prerequisites for `setVariable` to work?


mickaelistria:
I'm embedding this debug adapter (the vscode-node-debug2 more specifically, but I tracked down the issue to this component) in Eclipse IDE.
Most request work fine, however I tried to enable support for `setVariable`, which capability declare is enabled, but I face

```json
{"seq":93,"type":"event","event":"output","body":{"category":"telemetry","output":"ClientRequest/variables","data":{"Versions.DebugAdapterCore":"6.7.55","Versions.DebugAdapter":"1.33.0","Versions.Target.Version":"v10.16.3","successful":"true","timeTakenInMilliseconds":"1.438987","requestType":"request"}}}Content-Length: 385

{"seq":94,"type":"event","event":"output","body":{"category":"telemetry","output":"ClientRequest/setVariable","data":{"Versions.DebugAdapterCore":"6.7.55","Versions.DebugAdapter":"1.33.0","Versions.Target.Version":"v10.16.3","successful":"false","exceptionType":"firstChance","exceptionMessage":"The debug adapter doesn't recognize this command","timeTakenInMilliseconds":"0.318617"}}}Content-Length: 254

{"seq":95,"type":"response","request_seq":12,"command":"setVariable","success":false,"message":"[node-debug2] Unrecognized request: setVariable","body":{"error":{"id":1014,"format":"[node-debug2] Unrecognized request: setVariable","sendTelemetry":true}}}
```

Is there some other prerequisite to make `setVariable` work?

The js file I'm debugging is a base
```js
var n = 4;
console.log("hi" + n);
```
and I try to change value when breakpoint is on the 2nd line.

If I tweak the `chromeDebugAdapter.js` file to add
```js
    setVariable(args) {
	      this._variablesManager.setVariable(args);
    }
``` nearby `setBreakpoints` I then get the following error
```json
{"seq":96,"type":"event","event":"output","body":{"category":"telemetry","output":"ClientRequest/setVariable","data":{"Versions.DebugAdapterCore":"6.7.55","Versions.DebugAdapter":"1.33.0","Versions.Target.Version":"v10.16.3","successful":"true","timeTakenInMilliseconds":"4.713406","requestType":"request"}}}Content-Length: 1991

[...]

PR #516, opened 2019-08-27
re-add setVariable handler

EricCornelson:
fix missing setVariable hook for #514


mickaelistria:
This looks very similar to the hack I've successfully tried in JS file directly, so +1 from me.

PR #519, opened 2019-08-29
Single invalid source doesn't prevent using source map

digeff:
Single invalid source doesn't prevent using source map

We saw some source-map cases like sources: ["source.ts", null]

### DIFF
--- a/src/sourceMaps/sourceMap.ts
+++ b/src/sourceMaps/sourceMap.ts
@@ -22,6 +22,7 @@ import { isNotEmpty, isDefined, isNull } from '../chrome/utils/typedOperators';
 import { wrapWithMethodLogger } from '../chrome/logging/methodsCalledLogger';
 import { LocalizedError, registerGetLocalize } from '../chrome/utils/localization';
 import { SourceMapContents } from './sourceMapContents';
+import { telemetry } from '../telemetry';
 
 let localize = nls.loadMessageBundle();
 registerGetLocalize(() => localize = nls.loadMessageBundle());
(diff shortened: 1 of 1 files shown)


src/sourceMaps/sourceMap.ts
```diff
@@ -132,25 +133,14 @@ export class SourceMap {
...
-                // strip file://
-                return new LocalFileURL(sourcePath).filePathRepresentation;
-            }
-
-            if (!path.isAbsolute(sourcePath)) {
-                // Overrides not applied, use the computed sourceRoot
-                return path.resolve(computedSourceRoot, sourcePath);
-            } else {
-                return sourcePath;
+        const normalizedSources: string[] = [];
```
roblourens: Is it a problem that the indexes of the sources will change at this point if you just skip ones that fail? It looks like sourceMap.sources is referenced after this point and I can't tell if it's a problem that `sourceMap.sources.length !== normalizedSources.length`.

digeff: Yes, that will most likely be an issue. Good catch :)

Now we add the unormalized source when things fail:
                // We add the unnormalized source when things fail. Processing this source will most likely fail in other places of the code.
                // We need to insert something to preserve the indexes of the sources matching the ones for the mappings, etc...
                normalizedSources.push(`${sourcePath}`);

digeff:
Yes, I mistakenly thought it was sources...
I tried putting null on sourcesContent and everything worked. We don't have any code reading sourcesContent...

src/sourceMaps/sourceMap.ts
```diff
@@ -132,25 +133,18 @@ export class SourceMap {
...
+            try {
+                normalizedSources.push(SourceMap.normalizeSource(sourceMapPathOverrides, sourceMap, isVSClient, computedSourceRoot, sourcePath));
+            } catch (exception) {
+                // If we fail to process a single source path, we'll try to process the other ones anyways instead of just failing everything
+                // this is important because we've seen some sourcemap with null source entries
+                telemetry.reportError('Failed to parse sourcemap source', exception);
+
+                // We add the unnormalized source when things fail. Processing this source will most likely fail in other places of the code.
+                // We need to insert something to preserve the indexes of the sources matching the ones for the mappings, etc...
+                normalizedSources.push(`${sourcePath}`);
```
roblourens: Just `normalizedSources.push(sourcePath)`?

digeff: I want to convert it to a string in case it's null, or anything that is not a string...

roblourens: Oh ok

PR #521, opened 2019-08-30
Improve logging logic to avoid printting [object Object]

digeff:
Added some missing toString()
Created JSONObject which copies the properties of a JSON object and provides a custom toString() to improve logging

### DIFF
--- a/src/chrome/logging/json.ts
+++ b/src/chrome/logging/json.ts
@@ -1,9 +1,9 @@
 export class JSONObject {
-    private constructor(jsonProperties: object) {
+    private constructor(jsonProperties: unknown) {
         Object.assign(this, jsonProperties);
     }
 
-    public static create<T>(jsonProperties: object): JSONObject & T {
+    public static create<T>(jsonProperties: T): JSONObject & T {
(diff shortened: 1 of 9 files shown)


src/chrome/logging/json.ts
```diff
@@ -0,0 +1,17 @@
+export class JSONObject {
+    private constructor(jsonProperties: object) {
+        Object.assign(this, jsonProperties);
+    }
+
+    public static create<T>(jsonProperties: object): JSONObject & T {
+        return <JSONObject & T>new JSONObject(jsonProperties);
```
roblourens: Why do you need a static constructor that just calls `new`? Can the generic parameter go on the class?

digeff: We need the type of the Wrapped JSON object to be the same as the unwrapped type for things to type properly. "class JSONObject<T> implements T" is not valid. I could try to modify the return value of the constructor, which I think it's possible, but it's not "nice". So I opted to create a  static method with the whole purpose of retaining the T type of the JSON object.

I've just updated the types to reflect this more clearly

roblourens: I see thanks

PR #527, opened 2019-09-10
Now we don't send sourceReference if we can't retrieve the source

digeff:
Before this change we were sending sourceReference for sources we couldn't find the source for. This PR addresses that issue.
As an experiment, we implemented this change by inserting a function instead of a class like we normally do. We added some support to be able to do that.

* src/chrome/cdtpDebuggee/cdtpDIContainer.ts:
  - Exported getSourceTextRetrievability so it can be customized by chrome-debug

* src/chrome/client/sourceToClientConverter.ts:
  - Only include sourceReferences for sources we can actually get the source for...

* src/chrome/dependencyInjection.ts/bind.ts:
  - Don't modify injectables that are numbers

* src/chrome/dependencyInjection.ts/di.ts:
  - Add support for injecting functions instead of classes
  - Pass _componentCustomizationCallback to find places where we should've been passing it already

* src/chrome/dependencyInjection.ts/types.ts:
  - Add isValueComponent to identify if an identifier is a class or a value

* src/chrome/internal/sources/sourceTextRetriever.ts:
  - Refactor the text logic to create the getSourceTextRetrievability functions

* src/chrome/logging/methodsCalledLogger.ts:
  - Don't try to wrap numbers in a proxy

This change won't compile with a required change in chrome-debug: 
https://github.com/microsoft/vscode-chrome-debug/pull/930

### DIFF
--- a/src/chrome/internal/sources/sourceTextRetriever.ts
+++ b/src/chrome/internal/sources/sourceTextRetriever.ts
@@ -20,15 +20,39 @@ let localize = nls.loadMessageBundle();
 registerGetLocalize(() => localize = nls.loadMessageBundle());
 import { SourceContents } from './sourceContents';
 
+export interface IPossiblyRetrievableText {
+    isRetrievable: boolean;
+
+    retrieve(): Promise<SourceContents>;
+}
+
+export class RetrievableText implements IPossiblyRetrievableText {
+    public readonly isRetrievable = true;
+
+    public constructor(public readonly retrieve: () => Promise<SourceContents>) { }
+}
+
+export class NonRetrievableText implements IPossiblyRetrievableText {
+    public readonly isRetrievable = false;
+
+    public constructor(public readonly retrieve: () => never) { }
+}
+
+export interface ISourceTextRetriever {
+    text(source: ILoadedSource): Promise<SourceContents>;
+    retrievability(source: ILoadedSource): IPossiblyRetrievableText;
+}
+
 /**
  * Retrieves the text associated with a loaded source that maps to a JavaScript script file
  * (If the loaded source maps to an .html file or something different than a single script, a different class/API will need to be used)
  */
 @injectable()
-export class SourceTextRetriever {
+export class SourceTextRetriever implements ISourceTextRetriever {
     private _sourceToText = new ValidatedMap<ILoadedSource, Promise<SourceContents>>();
 
-    constructor(@inject(TYPES.IScriptSources) private readonly _scriptSources: IScriptSourcesRetriever) { }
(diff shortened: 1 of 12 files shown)


src/chrome/internal/sources/sourceTextRetriever.ts
```diff
@@ -20,47 +20,91 @@ let localize = nls.loadMessageBundle();
...
+    if (loadedSource.sourceScriptRelationship === SourceScriptRelationship.SourceIsSingleScript && scripts.length === 1) {
+        const singleScript = singleElementOfArray(scripts);
+        return new RetrievableText(() => scriptSources.getScriptSource(singleScript));
+    } else if (loadedSource.sourceScriptRelationship === SourceScriptRelationship.SourceIsSingleScript && scripts.length >= 2) {
+        /**
+         * We have two or more scripts associated with this source. At the moment we don't have any further support for this scenario
+         * so we just return the source of the first script. This won't be ideal if for some reason the source of both scripts
+         * isn't the same.
+         */
+        logger.warn(`${loadedSource} is associated with several ${printArray('scripts', scripts)} returning arbitrarily the source of the first one`);
```
roblourens: Maybe log the warning inside the RetrievableText callback so it shows up in the log next to where the text is returned?

src/chrome/internal/sources/sourceTextRetriever.ts
```diff
@@ -20,47 +20,91 @@ let localize = nls.loadMessageBundle();
...
+        const singleScript = scripts[0];
+        return new RetrievableText(() => scriptSources.getScriptSource(singleScript));
+    } else if (loadedSource.contentsLocation === ContentsLocation.PersistentStorage) {
+        // If this is a file, we don't want to cache it, so we return the contents immediately
+        return new RetrievableText(() => utils.readFileP(loadedSource.identifier.textRepresentation));
```
roblourens: When does the debug adapter have to return the text of a file that's on disk? Shouldn't vscode/VS always be able to read it?

src/chrome/internal/sources/sourceTextRetriever.ts
digeff: That makes sense

src/chrome/internal/sources/sourceTextRetriever.ts
digeff: As far as I know that's accurate. I implemented this because the protocol interface SourceArguments says:
   * Specifies the source content to load. Either source.path or source.sourceReference must be specified.
The source.path part made me think that someone might decide to do this with a path...

roblourens: Hm, I guess we technically need to handle it then.

roblourens:
Integration test failure?

PR #545, opened 2019-11-19
Added additional exception handling for environments that are not support getPossibleBreakpoints requests

### LINKED ISSUE #534: Add more exceptions handling for breakpoints position requests for inline breakpoints
Hello, I am a maintainer of [vscode-react-native](https://github.com/microsoft/vscode-react-native).
After the VS Code 1.39 update after setting a breakpoint when [debugging Hermes](https://github.com/microsoft/vscode-chrome-debug-core/issues/517) the following message shows up in debug console:
```
Error processing "breakpointLocations": TypeError: Cannot read property 'map' of undefined
    at Breakpoints.<anonymous> (c:\Work\vscode-react-native\node_modules\vscode-chrome-debug-core\out\src\chrome\breakpoints.js:303:68)
    at Generator.next (<anonymous>)
    at fulfilled (c:\Work\vscode-react-native\node_modules\vscode-chrome-debug-core\out\src\chrome\breakpoints.js:7:58)
```
I looked at the code and found out that `possibleBpResponse` could be `undefined` [at here](https://github.com/microsoft/vscode-chrome-debug-core/blob/31d19232b18071031a9d2489772fb68f1c333e5f/src/chrome/breakpoints.ts#L336) and this might be the cause of this message.
Logs for such case: [chrome-debug-core-session.txt](https://github.com/microsoft/vscode-chrome-debug-core/files/3728274/chrome-debug-core-session.txt)

Also, some Chrome DevTools implementations cannot handle breakpoints location requests correctly and throws errors due to this. Is it possible to disable these requests in some cases?

SounD120:
This PR forces `getBreakpointsLocation` function to swallow error messages in cases if environment doesn't support `getPossibleBreakpoints` request.
Fixes #534

### DIFF
--- a/src/chrome/breakpoints.ts
+++ b/src/chrome/breakpoints.ts
@@ -333,21 +333,27 @@ export class Breakpoints {
                     { scriptId: script.scriptId, lineNumber: args.endLine, columnNumber: args.endColumn || 0 } :
                     { scriptId: script.scriptId, lineNumber: args.line + 1, columnNumber: 0 };
 
-                const possibleBpResponse = await this.chrome.Debugger.getPossibleBreakpoints({
-                    start: { scriptId: script.scriptId, lineNumber: args.line, columnNumber: args.column || 0 },
-                    end,
-                    restrictToFunction: false
-                });
-                let breakpoints = possibleBpResponse.locations.map(loc => {
-                    return <DebugProtocol.Breakpoint>{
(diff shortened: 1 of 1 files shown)


SounD120:
@roblourens I changed check approach, and scoped this check down. Please let me know if have any concerns about that fix.

PR #546, opened 2019-11-22
Fix reference error issue when setting a new value for property

ahmadov:
Fix microsoft/vscode-node-debug2#250

### DIFF
--- a/src/chrome/variablesManager.ts
+++ b/src/chrome/variablesManager.ts
@@ -104,9 +104,10 @@ export class VariablesManager {
     }
 
     public setPropertyValue(objectId: string, propName: string, value: string): Promise<string> {
-        const setPropertyValueFn = `function() { return this["{propName}"] = {value} }`;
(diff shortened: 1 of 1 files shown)


roblourens:
FYI @EricCornelson it looks like the `$`s disappeared when moving this code to its own class.

ahmadov:
Yes, first I wanted to just add the missing `${}` but then I realized the type of the value argument is already a `string`. So, If we put the missing `$`, the function would be like that: `this[...] = ${value}` and putting `$` is not enough because if the value is `"foo"`  the result will be `this[...] = foo` which is also undefined.

The way to solve that issue is surrounding the `${value}` with (escape) double-quotes such as `this[...] = \"${value}\"` and in that case, it'd be always interpreted as strings.

I followed the other way because Chrome also uses the way "arguments". I think the underlying problem is type of the value is always a string.

ahmadov:
Hi @roblourens,
I got your point and changed the code back to `${}` as it was before.

PR #552, opened 2020-02-21
Fixed silent crash for cases when first child of grouped log is null

ArchDev:
Recently, one of apps I support started crashing the debugging process in extension and I find out that it happen because of null value returned as first message of console.startGroup. I would like to upstream my local patch as fix.


roblourens:
Thanks for the patch, but could you tell me how to reproduce this crash? `console.group(null)` seems to work fine for me

ArchDev:
I didn't dived deep in it, log was produced by Relay library and engine which app uses is Hermes. Not sure was this behaviour specific to Hermes or Relay just doing smth different. Is it important to figure out?

roblourens:
No that's ok, it's just useful to know that this happens with Hermes. I only test with Node/Chrome, Hermes has some slight differences.

PR #555, opened 2020-05-12
telemetry: prefix keys for classification in vscode

connor4312:
Opted to remap it inside the telemetry reporter so that this is a single change and doesn't break the exposed interface.

Once this is published and packages updated, https://github.com/microsoft/vscode/issues/97628 can be closed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2073,7 +2073,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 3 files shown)


roblourens:
Where are the !-prefixed keys filtered out? Will this cause any issues for VS people looking at telemetry if they want to query for old non-prefixed events and also the new events? @digeff

connor4312:
They're filtered in VS Code core, as of https://github.com/microsoft/vscode/pull/97627

An alternative approach would be to add some `_errorProperties` property to the log message, which would not require changing property names. I can make that change if we think that's better.

roblourens:
I think this is probably find but let's clear it with @digeff

digeff:
Assuming we get the telemetry properties in the backend, and nothing drops them because they start with !, then this is fine. Do you know if our backend supports an ! at the start of the properties?

connor4312:
@digeff I'm not sure what backend you guys use on the VS side but I see these properties come through correctly to the VS Code Kusto

digeff:
I think we use the same infrastructure, so I'd expect it to work in VS too.

Given that you already wrote this code, I'd merge it as is. I'll insert this into VS and see if I get the new properties. If I find any issues, we can implement the fix in a new PR. What do you think?

connor4312:
Sounds good

PR #559, opened 2020-11-04
Support urls without protocol

yepitschunked:
It's valid to have sourceMappingURLs without a protocol, e.g "sourceMappingUrl=//example.com/source.map". It's useful on pages that can be served over both http or https (common in development). This fixes the parser to accept this sort of URL as well.


roblourens:
Thanks for the PR, but I would consider this library deprecated, development has moved to https://github.com/microsoft/vscode-js-debug/. Please check whether the issue is fixed there.

